PR #2, opened 2018-09-27
Updated Card.java needs testing

smcdonald4812:
Added getters and constructors for this class, as well as boolean isAce for determining secondary value and rankValue based on ordinals. I did the coding here in GitHub, I'm not sure how to do the JUnit testing. Would someone be able to instruct me on how to do it?

### DIFF
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -10,18 +10,9 @@ package com.coderanch.blackjack;
 /**
  * A playing card used in a game of Blackjack.
  */
-final class Card {
-  
-  /**
-   * The rank of a {@link Card}.
-   * smcdonald4812 - filled in enum, created constructors, added an int rankValue, 
-   */
-  enum Rank {ONE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING, ACE};
+public final class Card {
 
-  /**
-   * The suit of a {@link Card}.
-   * smcdonald4812 - filled in enum
-   */
+  enum Rank {ONE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING, ACE};
   enum Suit {SPADES, HEARTS, CLUBS, DIAMONDS};
   
   // smcdonald4812 - value of the rank, secondary value for the ace, Rank, and Suit
@@ -32,9 +23,8 @@ final class Card {
   
   //smcdonald4812 - default constructor, non-null assignment.
   
-  Public Card() {
+  public Card() {
     super();
-    this.isAce = false;
     this.rankValue = 1;
     this.secondaryValue = 1;
     this.rank = Rank.ONE;
@@ -43,7 +33,7 @@ final class Card {
   
   //smcdonald4812 - Rank constructor must call this() as one has been provided.  .ordinal() is zerobased enum indexing.
   
-  Public Card(Rank rank) {
+  public Card(Rank rank) {
     this();
     this.rank = rank;
     this.rankValue += rank.ordinal();
@@ -62,8 +52,8 @@ final class Card {
   * smcdonald4812 - Suit, Rank constructor.
   */
   
-  Public Card(Rank rank, Suit suit) {
-    this(Rank rank);
+  public Card(Rank rank, Suit suit) {
+    this(rank);
     this.suit = suit;
   }
   
@@ -73,25 +63,19 @@ final class Card {
    * @return the rank of this card; never {@literal null}
    * smcdonald4812 - non-null in this(). There are no setters as Card shouldn't be changed once it is created.
    */
-  public Rank getRank() {
-    return this.rank;
-  }
-  
-  /**
-   * Gets the suit of this card.
-   * 
-   * @return the suit of this card; never {@literal null}
-   * smcdonald4812 - non-null in this().
-   */
-  
-  public Suit getSuit() {
+   
+   public Rank getRank() {
+	   return this.rank;
+   }
+   
+   public Suit getSuit() {
     return this.suit;
   }
   
   //smcdonald4812 - allows callers to get the card's rankValue. Non-null in this().
   
   public int getValue() {
-    return this.value;
+    return this.rankValue;
   }
   
   //smcdonald4812 - allows callers to get ace's secondary value, and if mistakenly used on other cards it gets their value.
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -16,7 +16,6 @@ public final class Card {
   enum Suit {SPADES, HEARTS, CLUBS, DIAMONDS};
   
   // smcdonald4812 - value of the rank, secondary value for the ace, Rank, and Suit
-  private int rankValue, secondaryValue;
   private Rank rank;
   private Suit suit;
   //smcdonald4812 - could add a String ID here as well with UUID if wanted.
@@ -25,8 +24,6 @@ public final class Card {
   
   public Card() {
     super();
-    this.rankValue = 1;
-    this.secondaryValue = 1;
     this.rank = Rank.ONE;
     this.suit = Suit.SPADES;
   }
@@ -36,16 +33,6 @@ public final class Card {
   public Card(Rank rank) {
     this();
     this.rank = rank;
-    this.rankValue += rank.ordinal();
-    this.secondaryValue += rank.ordinal();
-    if(rankValue > 10 && rankValue < 14) {
-      this.rankValue = 10;
-      this.secondaryValue = this.rankValue;
-    }
-    if(rankValue == 14) {
-      this.rankValue = 11;
-      this.secondaryValue = 1;
-    }
   }
   
   /*
@@ -65,22 +52,10 @@ public final class Card {
    */
    
    public Rank getRank() {
-	   return this.rank;
+    return this.rank;
    }
    
    public Suit getSuit() {
     return this.suit;
-  }
-  
-  //smcdonald4812 - allows callers to get the card's rankValue. Non-null in this().
-  
-  public int getValue() {
-    return this.rankValue;
-  }
-  
-  //smcdonald4812 - allows callers to get ace's secondary value, and if mistakenly used on other cards it gets their value.
-  
-  public int getSecondaryValue() {
-    return this.secondaryValue;
   }
 }
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -14,29 +14,89 @@ final class Card {
   
   /**
    * The rank of a {@link Card}.
+   * smcdonald4812 - filled in enum, created constructors, added an int rankValue, 
    */
-  enum Rank {}
+  enum Rank {ONE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING, ACE};
 
   /**
    * The suit of a {@link Card}.
+   * smcdonald4812 - filled in enum
    */
-  enum Suit {}
+  enum Suit {SPADES, HEARTS, CLUBS, DIAMONDS};
+  
+  // smcdonald4812 - value of the rank, secondary value for the ace, Rank, and Suit
+  private int rankValue, secondaryValue;
+  private Rank rank;
+  private Suit suit;
+  //smcdonald4812 - could add a String ID here as well with UUID if wanted.
+  
+  //smcdonald4812 - default constructor, non-null assignment.
+  
+  Public Card() {
+    super();
+    this.isAce = false;
+    this.rankValue = 1;
+    this.secondaryValue = 1;
+    this.rank = Rank.ONE;
+    this.suit = Suit.SPADES;
+  }
+  
+  //smcdonald4812 - Rank constructor must call this() as one has been provided.  .ordinal() is zerobased enum indexing.
+  
+  Public Card(Rank rank) {
+    this();
+    this.rank = rank;
+    this.rankValue += rank.ordinal();
+    this.secondaryValue += rank.ordinal();
+    if(rankValue > 10 && rankValue < 14) {
+      this.rankValue = 10;
+      this.secondaryValue = this.rankValue;
+    }
+    if(rankValue == 14) {
+      this.rankValue = 11;
+      this.secondaryValue = 1;
+    }
+  }
+  
+  /*
+  * smcdonald4812 - Suit, Rank constructor.
+  */
+  
+  Public Card(Rank rank, Suit suit) {
+    this(Rank rank);
+    this.suit = suit;
+  }
   
   /**
    * Gets the rank of this card.
    * 
    * @return the rank of this card; never {@literal null}
+   * smcdonald4812 - non-null in this(). There are no setters as Card shouldn't be changed once it is created.
    */
-  Rank rank() {
-    return null;
+  public Rank getRank() {
+    return this.rank;
   }
   
   /**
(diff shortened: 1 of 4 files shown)


chiquito83 (commented):
I'm thinking that maybe rankValue and secondaryValue should be left out as that's specific to a particular type of game, in case we later figure out that there is some type of Blackjack game that calculates it differently? Card class should not contain any logic I think, it should be just a simple immutable object. The logic could go to Table class (or similar)

smcdonald4812:
Fixed a couple dumb mistakes I made. Just saw your review and your edit @chiquito83, thanks for the fixes. I'm new at this, it's up to you guys how you want to do it. It's up to you guys if you want to get rid of the values, ordinals, and logic. I figured by only having getters and not setters the card can't be changed after construction. I added the logic because I thought calls like getRankValue() and getSecondaryValue() might make it easier inside a deck or table class, and this way the card's values can't be changed after it's created.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
-   * The rank of a {@link Card}.
-   */
-  enum Rank {}
+public final class Card {
 
-  /**
-   * The suit of a {@link Card}.
-   */
-  enum Suit {}
+  enum Rank {ONE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING, ACE};
```
nibsi: Playing cards don't have the rank ONE. I suggest to move ACE to the start of the list.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+  enum Suit {SPADES, HEARTS, CLUBS, DIAMONDS};
```
nibsi: Always put at least one space around curly braces.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
   
-  /**
-   * Gets the rank of this card.
-   * 
-   * @return the rank of this card; never {@literal null}
-   */
-  Rank rank() {
-    return null;
+  // smcdonald4812 - value of the rank, secondary value for the ace, Rank, and Suit
```
nibsi: Comment is unnecessary.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
 /**
  * A playing card used in a game of Blackjack.
  */
-final class Card {
-  
-  /**
-   * The rank of a {@link Card}.
-   */
-  enum Rank {}
+public final class Card {
```
nibsi: Don't make declarations public until necessary.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
   
-  /**
-   * Gets the rank of this card.
-   * 
-   * @return the rank of this card; never {@literal null}
-   */
-  Rank rank() {
-    return null;
+  // smcdonald4812 - value of the rank, secondary value for the ace, Rank, and Suit
+  private Rank rank;
```
nibsi: Make fields final unless the class is intended to be mutable.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+  private Suit suit;
+  //smcdonald4812 - could add a String ID here as well with UUID if wanted.
```
nibsi: IDs are rarely useful for value types.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+  
+  //smcdonald4812 - default constructor, non-null assignment.
+  
+  public Card() {
```
nibsi: Having the one (ace?) of spades be the default seems rather arbitrary. I suggest removing the parameterless constructor so that the caller is always required to pass a rand and suit explicitly.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+    super();
+    this.rank = Rank.ONE;
+    this.suit = Suit.SPADES;
+  }
+  
+  //smcdonald4812 - Rank constructor must call this() as one has been provided.  .ordinal() is zerobased enum indexing.
+  
+  public Card(Rank rank) {
```
nibsi: Same as parameterless constructor. Always require passing both rank and suit.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+    this.rank = rank;
+  }
+  
+  /*
+  * smcdonald4812 - Suit, Rank constructor.
+  */
+  
+  public Card(Rank rank, Suit suit) {
+    this(rank);
+    this.suit = suit;
```
nibsi: Parameters aren't checked.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+  public Card(Rank rank) {
+    this();
+    this.rank = rank;
+  }
+  
+  /*
+  * smcdonald4812 - Suit, Rank constructor.
+  */
+  
+  public Card(Rank rank, Suit suit) {
```
nibsi, line 37: Write Javadoc for all members.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -10,33 +10,52 @@
...
+   * Gets the rank of this card.
    * 
-   * @return the suit of this card; never {@literal null}
+   * @return the rank of this card; never {@literal null}
+   * smcdonald4812 - non-null in this(). There are no setters as Card shouldn't be changed once it is created.
    */
-  Suit suit() {
-    return null;
+   
+   public Rank getRank() {
```
nibsi, line 47: Any reason for this rename? It's not a problem, I'm just curious.

nibsi (changes requested):

core/src/main/java/com/coderanch/blackjack/Card.java
liutaurasvilda: After no-args (it isn't default, default is being added by Java when no constructor is defined explicitly) constructor remove, rank and suit fields can become final.

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: Is that the two asterisk comment /** **/ ?

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: I've been studying for the OCA, and it stats that all getters and setters must be either get, set, or is. I just thought that was standard practice?

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: I figured that ONE would be used in the no-arg constructor as a useless value, just to ensure that Card suit and rank can't be null. I see that there should be no other constructor other than Card(Rank rank, Suit suit) now. I also noticed the addition of a utility class in the project? What is a utility class, is it like a singleton?

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -12,31 +12,66 @@
...
+    this.suit = getConstSuit(suit);
+  }
+  
+  /*
+  * Oracle states that private methods should not have a JavaDoc, 
+  * also that assertions should not be used for runtime exceptions.
+  * I have implemented these getters as I;ve been told logic should not be placed
+  * in the constructor itself. I'm not sure if it is okay to do it this way? 
+  * Any feedback would be geat. Also, suit not being of Suit is an error (cannot find symbol).
+  * How is this supposed to be checked for without surrounding constructor in try catch?
```
nibsi: It's good practice to write Javadoc for private methods and constructors anyway, because people other than yourself may want to call them and they want to know what those operations are supposed to do.

You can put logic in the constructor of that logic is necessary to initialize the object. When people say not to put logic in the constructor, they mean additional operations besides parameter checking and field initialization.

Please don't put questions about the code in the code. Push the code and then annotate it in GitHub, like I'm doing right now.

nibsi: Instead of calling these methods to return a default value, throw an exception when a parameter is invalid. One of the unit tests actually tests this, so you will have written the correct solution when the tests pass.

How do you compile the code? Do you have Maven installed on your system? Are you using an IDE?

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -12,31 +12,66 @@
...
+    }
+    return suit;
+  }
+  
+  private Rank getConstRank(Rank rank) {
+    if(rank == null) {
+      return Rank.TWO;
+    }
+    return rank;
+  }
```
nibsi, line 40: Once you throw exception from your constructor, these methods become unnecessary. Even if they're not, you should have made them static because they don't rely on instance members.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: It's widely practiced, but I wouldn't say that accessors **must** follow this convention. Personally I prefer to use a noun without the `get`- prefix if the getter represents a simple property of the object that just returns a field without any complex logic.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -12,31 +12,66 @@
  */
 final class Card {
   
+  enum Rank { ACE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING };
+  enum Suit { SPADES, HEARTS, CLUBS, DIAMONDS };
```
nibsi: The Javadoc on these enums disappeared.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -12,31 +12,66 @@
...
+   
+   public Rank getRank() {
+    return this.rank;
+   }
+   
   /**
    * Gets the suit of this card.
    * 
    * @return the suit of this card; never {@literal null}
+   *
```
nibsi: Why the additional newlines?

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: @Nibsi I normally write code in notepad++, and then I use the command line to javac and test with either another class or I add code to it for output. I have Eclipse and NetBeans, but I've been practicing for the OCA, so I've been trying to do it the hard way to learn more. Honestly, I didn't realize how bad I was at writing code until I started doing it this way. Although, this last time I edited the class here on the GitHub website.

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: I didn't know that, should methods always be static when they don't rely on members?

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: I was following the guidelines I found online. I've never created a JavaDoc before.

core/src/main/java/com/coderanch/blackjack/Card.java
smcdonald4812: That makes sense. I just thought it was one of those things like camelCase.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: If you don't intend for them to be inherited, yes.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: Yep.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -13,48 +13,46 @@
...
+   * @param suit (required) Suit value. Must have value of enum Suit.
+   */
   public Card(Rank rank, Suit suit) {
-    if (rank == null || suit == null) {
-      throw new IllegalArgumentException("Rank and suit should not be null");
-    }
-    this.rank = rank;
-    this.suit = suit;
+    this.rank = Objects.requireNonNull(rank, "Rank should not be null");
+    this.suit = Objects.requireNonNull(suit, "Suit should not be null");
```
nibsi, line 39: This change causes the tests to fail because they expect `IllegalArgumentException`. We need to decide what exception type we want to use for null-checks in method/constructor parameters. If we do want to stick with `IllegalArgumentException`, we could include my Require library to make this process less painful.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -13,48 +13,46 @@
...
-   
+
   /**
    * Gets the suit of this card.
+   *
    * @return the suit of this card; never {@literal null}
    */
-  
    public Suit getSuit() {
     return this.suit;
```
nibsi, line 57: Wrong indentation on the method bodies of `getRank()` and `getSuit()`.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: For now, I suggest we change the tests to expect `NullPointerException`, we can always change it later.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: @smcdonald4812 It's fine to use Notepad++, but since this is a Maven project, I suggest you download and install Maven and then compile the code using `mvn package`. That way the tests are run and you can see what changes you have to make before you push a broken build.

core/src/main/java/com/coderanch/blackjack/Card.java
liutaurasvilda: Sorry, I wasn't aware about such existing test. I need to get more familiar with where this project headed to already. I was researching some time ago which exception is more appropriate to throw when unwanted null is passed as a parameter: either NullPointer or IllegalArgument exception, and came to conclusion, that IllegalArgumentException is more appropriate when there is non-null unwanted parameter value, and NullPointerException when there is actually a null, so you know right away what you are dealing with. What is your opinion about that?

core/src/main/java/com/coderanch/blackjack/Card.java
liutaurasvilda: It seems everybody uses 2 spaces. Is it what we agreed on? So I left them unchanged, fixed only javadocs formatting.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: I prefer the option that gives more fine grained information about what went wrong. A `NullPointerException` to me implies that you dereferenced null, not necessarily because you supplied a wrong argument to a method. Throwing `NullPointerException` without actually dereferencing the value hides that the problem was caused because you passed a wrong argument. I find that I care that a wrong argument was passed, not necessarily that that argument was null or something else.

I don't have a very strong preference, but I find it easier to debug an application when using `IllegalArgumentException`.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: Okay, will fix later.

PR #4, opened 2018-09-30
Update .gitignore

liutaurasvilda:
To ignore IntelliJ IDEA config files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 target/
+.idea/
+*.iml


chiquito83 (approved):

.gitignore
```diff
@@ -1 +1,3 @@
-target/
\ No newline at end of file
+target/
+.idea/
+*.iml
```
nibsi, line 3: Are these personal settings? If they are project files that are relevant for everyone using IDEA, then they actually do belong in the repository.

A Google search indicates that for Maven projects, you need to include all files in the `.idea/` folder except `workspace.xml`, `usage.statistics.xml`, and `tasks.xml`.

liutaurasvilda: Yes, they are personal settings. When you import project, whether it is Maven or just simply plain project, IntelliJ creates bunch of files which most of them reside in .idea directory and two *.iml files which are also IntelliJ specific files. Mostly they describe which language level you have set in your (note: your) project settings, which libraries versions, dir mappings, and other metadata. Important is, that you don't need to have them, your IDE creates all them when you first time import project and then maintain its settings during project changes. When there are different people with different IDE's, they are just noise in my opinion to upload to repository. At least I never saw uploaded them (for instance at my work).

nibsi: Okay!

nibsi (approved):

PR #6, opened 2018-10-01
Whitelist branches to build on Travis CI

nibsi:
By default, any push to a branch will trigger continuous integration.
Whitelisting branches causes only those branches to be built by Travis.
This doesn't affect pull request updates, which are always built.


liutaurasvilda (approved):

liutaurasvilda (approved):

liutaurasvilda:
Ok, checked whether the same person (myself in this case) can be counted as 2 reviews. Nope - as expected, so we can't abuse it, weird that it let me approve it if I did so already though.

nibsi:
Yeah not sure why it would let you approve twice when nothing has changed.

ksnortum:
I'd review this, but I don't know how to test the CI.

nibsi:
Well, reviews are mostly about whether you can find things that are obviously wrong. It could have been that I made a typo in the regex or something, or used a construct that isn't valid for the Ruby flavor. That, and it's a good way to become familiar with part of the codebase.

I'm completely new to Travis CI as well. Most of what I'm doing here are try outs. The upside of this stuff being new is that we don't have much to break.

nibsi:
Maybe I can explain a bit better why I made this change. Before, Travis would build and test our code for every push on every branch, as well as every update to a pull request. While we certainly want pull requests to be tested thoroughly, testing other commits is only really useful on long living branches that are supposed to be stable at all times.

I whitelisted the master branch, plus all release/ branches, because we want to make sure those are stable.

chiquito83 (approved):

PR #10, opened 2018-10-14
Created cli sub module and added UserInterface and InputUtilities classes to It

Ganesh-Patekar:
InputUtilities for validating user inputs. UserInterface will be updated as soon as I add Hand as well as Player class.


cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (C) 2018 Coderanch.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.coderanch.blackjack;
```
nibsi, line 8: The user interface should not interact with internal parts of the model, therefore they should be in different packages.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Scanner;
+
+/**
+ * Utility class to verify input values.
+ */
+final class InputUtilities {
+	/**
+	 * This method verifies If name is valid Or not. Valid name consists of
```
nibsi, line 17: Mind use of capitalization in your documentation. 'If' and 'or' should be written in lowercase.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+	 * characters and spaces having minimum length 3 and maximum 20.
```
nibsi, line 18: Why the arbitrary sizes? What's wrong with a name of one character? What's wrong with a name of 21 characters? Why is a name that consists of only whitespace valid?

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+	 * 
+	 * @param name
+	 *            Name to be validated. false If passed name is invalid name.
```
nibsi, line 21: Don't put descriptions of the return value with the parameter description. That's what the return tag is for.

nibsi: Javadoc convention is to start the first sentence of parameter and return value descriptions with a lower case letter, *unless* it's a full sentence. Example: "the name to be validated." versus "This parameter represents the name to be validated.". The first form is preferred.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+	 * @return Returns true If It's a valid name else return false.
```
nibsi, line 22: Give a detailed descriptor here of what it means for the name to be valid.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+	 */
+	public static boolean isValidName(String name) {
+
+		if (name.length() < 3 || name.length() > 20) {
```
nibsi, line 26: No null check.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+			return false;
+		}
+		name = name.toUpperCase();
+		for (int i = 0; i < name.length(); i++) {
+
+			char ch = name.charAt(i);
+
+			if ((!(ch >= 65 && ch <= 90)) && (ch != ' ')) {
```
nibsi, line 34: You didn't describe this requirement in the documentation. Why am I not allowed to use a name containing Chinese ideograms, for instance?

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+	 *            Name to be validated. false If passed name is invalid name.
+	 * @return Returns true If It's a valid name else return false.
+	 */
+	public static boolean isValidName(String name) {
+
+		if (name.length() < 3 || name.length() > 20) {
+			return false;
+		}
+		name = name.toUpperCase();
+		for (int i = 0; i < name.length(); i++) {
```
nibsi, line 30: You're doing a lot of work that can be replaced by a simple regex match.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+		}
+		return true;
+	}
+
+	/**
+	 * This method verifies whether users input is whole number Or not.
+	 * 
+	 * @return Returns -1 If user input is not an int value.
+	 */
+	public static int isWholeNumber() {
```
nibsi, line 47: I doubt this method is very useful. The caller presumably already uses a scanner, and therefore only had to call `hasNextInt()`.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+		Scanner sc = new Scanner(System.in);
+		return sc.hasNextInt() ? sc.nextInt() : -1;
```
nibsi, line 49: Don't return magic values from methods to indicate exceptional situations. Use exceptions, `Optional` or a custom type. In this case you don't need this method at all though.

cli/src/main/java/com/coderanch/blackjack/UserInterface.java
```diff
@@ -0,0 +1,71 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.coderanch.blackjack;
+
+import java.util.Scanner;
+
+final class UserInterface {
+
+	private void mainMenu() {
+		Scanner sc = new Scanner(System.in);
```
nibsi, line 15: Don't look up static resources. Inject them with method parameters. That means passing `System.in` through a parameter that takes an `InputStream`.

nibsi: Better yet, pass an InputStream and a PrintWriter as arguments to the constructor, seeing as your method aren't static anyway.

cli/src/main/java/com/coderanch/blackjack/UserInterface.java
```diff
@@ -0,0 +1,71 @@
...
+		int choice;
+		String mainMenuStr = "1. Play Game \n10. Exit\nEnter choice";
+		do {
+			System.out.println(mainMenuStr);
+			choice = InputUtilities.isWholeNumber();
```
nibsi, line 20: It's really counterintuitive that a method with a name starting with 'is' would return an int.

Besides, it's absolutely unnecessary for choice to be a number, because you're not going to be performing numeric operations. Just leave it a string and compare it to string constants.

cli/src/main/java/com/coderanch/blackjack/UserInterface.java
```diff
@@ -0,0 +1,71 @@
...
+			case 1:
+				System.out.println("Enter balance");
+				int balance = InputUtilities.isWholeNumber();
+				if (balance == -1) {
+					System.out.println("Invalid balance, only whole number is allowed");
+					break;
+				}
+				// Rest code here
+				break;
+			case 2:
```
nibsi, line 53: Why add cases when you can't handle them yet? Remember YAGNI.

pom.xml
```diff
@@ -36,6 +36,7 @@
 
     <modules>
         <module>core</module>
+        <module>cli</module>
```
nibsi: You forgot to add a POM for your cli module.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Okay, will keep in mind.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Agreed there is nothing wrong with a name of one character since It's just a game. I wrote condition keeping in mind that name should look real the way we have on our forum although agreed there could be a name with two characters. Not sure about one character. 

Ya name can be of more than 21 characters.  Even though we can have String of any length do you think we shouldn't have condition for maximum length for name? I mean what If someone types name of 100 characters which may not cause any problem but doesn't seem legit name.

I think I'm getting over concerned about name must appear legit or real one that's why allowed only white spaces.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: You mean all important .java classes of internal parts of the model will be in com.coderanch.blackjack package? 
will it be okay if i create package com.coderanch.ui where UserInterface class will reside ?

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Updated locally.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Done.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Oh! actually i checked null condition for name in the RegularPlayer class's constructor which suppose to be checked in this method.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: That's true if user of this application is Chinese he/she may use Chinese ideogram, then if we remove condition of characters then i guess condition for name will be only null check and min length i.e. name at least be of a character. Hope i am correct?

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Yes I got that, I saw some other codes on our forum which uses regex and also your code of Card and Cards which uses stream. 

Now I'm at least okay with Git commands, Git hub, Maven commands ( almost done with necessary commands, how It works, will be done by tomorrow ) 

I've listed down the things I need to learn i.e. JUnit5 testing, regex, Lambdas, Stream API and most importantly OOP's way of designing maintainable application.
 
I need to start getting into those, which I'm already, a bit slow but will cover all for sure.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Updated. removed unnecessary method.

cli/src/main/java/com/coderanch/blackjack/UserInterface.java
Ganesh-Patekar: YAGNI ?

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
nibsi: Would that package name make sense considering that CodeRanch might write UIs for different applications? Or that it might even write multiple different UIs for Blackjack?

I propose `com.coderanch.blackjack.cli`.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
nibsi: You'll be surprised what are valid names in different cultures. Some cultures use names that can't even be expressed with characters we have in Unicode.

I find the only limitation you will ever want to place on a written name is that it shouldn't start or end with white space, and it should not be empty.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: Ah! I see, that way. Got now.

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
Ganesh-Patekar: That's interesting to know. Will put those conditions then ...

ksnortum:
You Ain't Gonna Need It

---
Knute Snortum
(via Gmail)

cli/src/main/java/com/coderanch/blackjack/InputUtilities.java
nibsi: Also check that the name is trimmed, i.e. doesn't start or end with white space.

cli/src/main/java/com/coderanch/blackjack/UserInterface.java
nibsi: As Knute has mentioned, You Ain't Gonna Need It. Don't write code until you need it, because most of the time it will end up being unused anyway.

pom.xml
nibsi: I fixed this.

nibsi:
@Ganesh-Patekar I merged some changes into this branch. How are you doing?

nibsi:
It appears this pull request has been abandoned. I'm closing it. You can create a new pull request when you want to continue.

PR #19, opened 2021-03-17
add hand class:

### LINKED ISSUE #21: Model a hand of cards
Write classes to model a hand of playing cards.

Add a method that determines the greatest total point value of a hand of cards that is less than or equal the maximum allowed point value. If no such value can be determined, it returns the smallest total point value of the hand of cards.

aquino-a:
-receives card
 -calculates score and best score
 -test scoring

### DIFF
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -39,7 +39,7 @@ final class Card implements Comparable<Card> {
         QUEEN (10),
         KING  (10);
 
-        private int points;
+        private final int points;
 
         Rank(int points){
             this.points = points;
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -44,10 +44,20 @@ final class Card implements Comparable<Card> {
          */
         private final int points;
 
+        /**
+         * Contructs a Rank enum with the given value.
+         *
+         * @param points the points the rank is worth.
+         */
         Rank(int points) {
             this.points = points;
         }
 
+        /**
+         * Get the point value of the rank.
+         *
+         * @return the point value of the rank.
+         */
         int points() {
             return points;
         }
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -102,7 +102,6 @@ final class Card implements Comparable<Card> {
 
     /**
      * Compares this card to another card.
-     * <p>
      * Cards are first compared by rank, then by suit.
      *
      * @param other {@inheritDoc}
@@ -115,7 +114,6 @@ final class Card implements Comparable<Card> {
 
     /**
      * Compares this card to another object for equality.
-     * <p>
      * Cards are considered equal if they have the same rank and suit.
      *
      * @param object {@inheritDoc}
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -100,10 +100,6 @@ final class Card implements Comparable<Card> {
         return this.suit;
     }
 
-    int points() {
-        return this.rank.points;
-    }
-
     /**
      * Compares this card to another card.
      * Cards are first compared by rank, then by suit.
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -25,13 +25,13 @@ final class Card implements Comparable<Card> {
      * The rank of a card.
      */
     enum Rank {
-        ACE(1,11), TWO(2), THREE(3), FOUR(4), FIVE(5),
+        ACE(11), TWO(2), THREE(3), FOUR(4), FIVE(5),
         SIX(6), SEVEN(7), EIGHT(8), NINE(9),
         TEN(10), JACK(10), QUEEN(10), KING(10);
 
-        private int[] points;
+        private int points;
 
-        Rank(int... points){
+        Rank(int points){
             this.points = points;
         }
     }
@@ -77,7 +77,7 @@ final class Card implements Comparable<Card> {
         return this.suit;
     }
 
-    int[] values() {
+    int points() {
         return this.rank.points;
     }
 
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -25,9 +25,19 @@ final class Card implements Comparable<Card> {
      * The rank of a card.
      */
     enum Rank {
-        ACE(11), TWO(2), THREE(3), FOUR(4), FIVE(5),
-        SIX(6), SEVEN(7), EIGHT(8), NINE(9),
-        TEN(10), JACK(10), QUEEN(10), KING(10);
+        ACE   (11),
+        TWO   ( 2),
+        THREE ( 3),
+        FOUR  ( 4),
+        FIVE  ( 5),
+        SIX   ( 6),
+        SEVEN ( 7),
+        EIGHT ( 8),
+        NINE  ( 9),
+        TEN   (10),
+        JACK  (10),
+        QUEEN (10),
+        KING  (10);
 
         private int points;
 
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -25,7 +25,15 @@ final class Card implements Comparable<Card> {
      * The rank of a card.
      */
     enum Rank {
-        ACE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING
+        ACE(1,11), TWO(2), THREE(3), FOUR(4), FIVE(5),
+        SIX(6), SEVEN(7), EIGHT(8), NINE(9),
+        TEN(10), JACK(10), QUEEN(10), KING(10);
+
+        public int[] values;
+
+        private Rank(int...values){
+            this.values = values;
+        }
     }
 
     /**
@@ -68,6 +76,8 @@ final class Card implements Comparable<Card> {
     Suit suit() {
         return this.suit;
     }
+
+    int[] values() { return this.rank.values; }
 
     /**
      * Compares this card to another card.
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -29,10 +29,10 @@ final class Card implements Comparable<Card> {
         SIX(6), SEVEN(7), EIGHT(8), NINE(9),
         TEN(10), JACK(10), QUEEN(10), KING(10);
 
-        public int[] values;
+        private int[] points;
 
-        private Rank(int...values){
-            this.values = values;
+        Rank(int... points){
+            this.points = points;
         }
     }
 
@@ -77,7 +77,7 @@ final class Card implements Comparable<Card> {
         return this.suit;
     }
 
-    int[] values() { return this.rank.values; }
+    int[] values() { return this.rank.points; }
 
     /**
      * Compares this card to another card.
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -75,6 +75,7 @@ final class Card implements Comparable<Card> {
      *
      * @param rank the rank of the new card.
      * @param suit the suit of the new card.
+     *
      * @throws IllegalArgumentException if either {@code rank} or {@code suit} is {@code null}.
      */
     Card(Rank rank, Suit suit) {
@@ -102,9 +103,11 @@ final class Card implements Comparable<Card> {
 
     /**
      * Compares this card to another card.
+     *
      * Cards are first compared by rank, then by suit.
      *
      * @param other {@inheritDoc}
+     *
      * @return {@inheritDoc}
      */
     @Override
@@ -114,9 +117,11 @@ final class Card implements Comparable<Card> {
 
     /**
      * Compares this card to another object for equality.
+     *
      * Cards are considered equal if they have the same rank and suit.
      *
      * @param object {@inheritDoc}
+     *
      * @return {@inheritDoc}
      */
     @Override
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -48,7 +48,7 @@ final class Card implements Comparable<Card> {
             this.points = points;
         }
 
-        public int getPoints() {
+        int points() {
             return points;
         }
     }
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -47,6 +47,10 @@ final class Card implements Comparable<Card> {
         Rank(int points) {
             this.points = points;
         }
+
+        public int getPoints() {
+            return points;
+        }
     }
 
     /**
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -25,23 +25,26 @@ final class Card implements Comparable<Card> {
      * The rank of a card.
      */
     enum Rank {
-        ACE   (11),
-        TWO   ( 2),
-        THREE ( 3),
-        FOUR  ( 4),
-        FIVE  ( 5),
-        SIX   ( 6),
-        SEVEN ( 7),
-        EIGHT ( 8),
-        NINE  ( 9),
-        TEN   (10),
-        JACK  (10),
-        QUEEN (10),
-        KING  (10);
+        ACE(11),
+        TWO(2),
+        THREE(3),
+        FOUR(4),
+        FIVE(5),
+        SIX(6),
+        SEVEN(7),
+        EIGHT(8),
+        NINE(9),
+        TEN(10),
+        JACK(10),
+        QUEEN(10),
+        KING(10);
 
+        /**
+         * The points value of the rank.
+         */
         private final int points;
 
-        Rank(int points){
+        Rank(int points) {
             this.points = points;
         }
     }
@@ -68,7 +71,6 @@ final class Card implements Comparable<Card> {
      *
      * @param rank the rank of the new card.
      * @param suit the suit of the new card.
-     *
      * @throws IllegalArgumentException if either {@code rank} or {@code suit} is {@code null}.
      */
     Card(Rank rank, Suit suit) {
@@ -100,11 +102,10 @@ final class Card implements Comparable<Card> {
 
     /**
      * Compares this card to another card.
-     *
+     * <p>
      * Cards are first compared by rank, then by suit.
      *
      * @param other {@inheritDoc}
-     *
      * @return {@inheritDoc}
      */
     @Override
@@ -114,11 +115,10 @@ final class Card implements Comparable<Card> {
 
     /**
      * Compares this card to another object for equality.
-     *
+     * <p>
      * Cards are considered equal if they have the same rank and suit.
      *
      * @param object {@inheritDoc}
-     *
      * @return {@inheritDoc}
      */
     @Override
--- a/core/src/main/java/com/coderanch/blackjack/Card.java
+++ b/core/src/main/java/com/coderanch/blackjack/Card.java
@@ -77,7 +77,9 @@ final class Card implements Comparable<Card> {
         return this.suit;
     }
 
-    int[] values() { return this.rank.points; }
+    int[] values() {
+        return this.rank.points;
+    }
 
     /**
      * Compares this card to another card.
--- a/core/src/main/java/com/coderanch/blackjack/Hand.java
+++ b/core/src/main/java/com/coderanch/blackjack/Hand.java
@@ -1,3 +1,10 @@
+/*
+ * Copyright (C) 2018 Coderanch.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
 package com.coderanch.blackjack;
 
 import java.util.*;
@@ -65,7 +72,9 @@ public class Hand {
      *
      * @return the current best score.
      */
-    public int bestScore() { return this.bestScore; }
+    public int bestScore() {
+        return this.bestScore;
+    }
 
     /**
      * Stops the hand.  Cannot receive more cards.
--- a/core/src/main/java/com/coderanch/blackjack/Hand.java
+++ b/core/src/main/java/com/coderanch/blackjack/Hand.java
@@ -58,7 +58,7 @@ final class Hand {
         return new Hand(newCards);
     }
 
-    private List<Integer> calculateScores() {
+    private static List<Integer> calculateScores(List<Card> cards) {
         var startingScore = cards.stream()
                 .filter(c -> c.rank() != Card.Rank.ACE)
                 .mapToInt(Card::points)
@@ -82,7 +82,7 @@ final class Hand {
         return scores;
     }
 
-    private int calculateBestScore(List<Integer> possibleScores) {
+    private static int calculateBestScore(List<Integer> possibleScores) {
         return possibleScores.stream()
                 .sorted((o1, o2) -> Integer.compare(o2, o1))
                 .filter(i -> i <= MAX_LEGAL_SCORE)
@@ -96,7 +96,7 @@ final class Hand {
      * @return the best score.
      */
     public int bestScore() {
-        var scores = calculateScores();
+        var scores = calculateScores(this.cards);
         return calculateBestScore(scores);
     }
 }
--- a/core/src/main/java/com/coderanch/blackjack/Hand.java
+++ b/core/src/main/java/com/coderanch/blackjack/Hand.java
@@ -23,6 +23,9 @@ final class Hand {
      */
     private static final int MAX_LEGAL_SCORE = 21;
 
+    /**
+     * The cards in the hand.
+     */
     private final List<Card> cards;
 
     /**
@@ -32,7 +35,7 @@ final class Hand {
      * @param card2 second card dealt
      * @throws NullPointerException if either {@code card} or {@code card2} is {@code null}.
      */
-    public Hand(Card card, Card card2) {
+    Hand(Card card, Card card2) {
         this(List.of(card, card2));
     }
 
--- a/core/src/main/java/com/coderanch/blackjack/Hand.java
+++ b/core/src/main/java/com/coderanch/blackjack/Hand.java
@@ -24,8 +24,6 @@ final class Hand {
     private static final int MAX_LEGAL_SCORE = 21;
 
     private final List<Card> cards;
-    private List<Integer> possibleScores;
-    private int bestScore;
 
     /**
      * Creates a new hand.
@@ -40,8 +38,6 @@ final class Hand {
 
     private Hand(List<Card> cards) {
         this.cards = cards;
-        this.possibleScores = calculateScores();
-        this.bestScore = calculateBestScore();
     }
 
     /**
@@ -83,8 +79,8 @@ final class Hand {
         return scores;
     }
 
-    private int calculateBestScore() {
-        return this.possibleScores.stream()
+    private int calculateBestScore(List<Integer> possibleScores) {
+        return possibleScores.stream()
                 .sorted((o1, o2) -> Integer.compare(o2, o1))
                 .filter(i -> i <= MAX_LEGAL_SCORE)
                 .findFirst()
@@ -97,6 +93,7 @@ final class Hand {
      * @return the best score.
      */
     public int bestScore() {
-        return this.bestScore;
+        var scores = calculateScores();
+        return calculateBestScore(scores);
     }
 }
--- a/core/src/main/java/com/coderanch/blackjack/Hand.java
+++ b/core/src/main/java/com/coderanch/blackjack/Hand.java
@@ -9,6 +9,7 @@ package com.coderanch.blackjack;
 
 import java.util.ArrayList;
 import java.util.List;
+
 import static java.util.function.Predicate.not;
 
 import static com.coderanch.blackjack.Card.Rank.ACE;
@@ -25,7 +26,7 @@ final class Hand {
     /**
      * Highest score in Blackjack.
      */
-    private static final int MAX_LEGAL_SCORE = 21;
+    static final int MAX_LEGAL_SCORE = 21;
 
     /**
      * The cards in the hand.
@@ -87,26 +88,26 @@ final class Hand {
             .filter(ACE::equals)
             .count();
 
-        var bestScore = calculateBestScore(minimumScore, numberOfFreeAces);
-        if (bestScore > MAX_LEGAL_SCORE) {
-            return 0;
-        }
-        else {
-            return bestScore;
-        }
+        return calculateBestScore(minimumScore, numberOfFreeAces);
     }
 
     private static int calculateBestScore(int minimumScore, int numberOfFreeAces) {
-        if (numberOfFreeAces <= 0) {
+        if (numberOfFreeAces == 0) {
             return minimumScore;
         }
         var bestScoreWithBigAce = calculateBestScore(minimumScore + ACE.points(), numberOfFreeAces - 1);
 
-        if (bestScoreWithBigAce > MAX_LEGAL_SCORE) {
-            return calculateBestScore(minimumScore + 1, numberOfFreeAces - 1);
-        }
-        else {
-            return bestScoreWithBigAce;
-        }
+        return bestScoreWithBigAce > MAX_LEGAL_SCORE
+            ? calculateBestScore(minimumScore + 1, numberOfFreeAces - 1)
+            : bestScoreWithBigAce;
+    }
+
+    /**
+     * Get whether the hand is bust or not.
+     *
+     * @return whether the hand is bust or not.
+     */
+    boolean isBust() {
+        return bestScore() > MAX_LEGAL_SCORE;
     }
 }
--- a/core/src/main/java/com/coderanch/blackjack/Hand.java
+++ b/core/src/main/java/com/coderanch/blackjack/Hand.java
@@ -9,7 +9,9 @@ package com.coderanch.blackjack;
 
 import java.util.*;
 
+import static com.coderanch.blackjack.Card.Rank.ACE;
 import static com.coderanch.util.require.Require.requireThat;
+import static java.util.function.Predicate.not;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
 
@@ -37,7 +39,7 @@ final class Hand {
      */
     Hand(Card firstCard, Card secondCard) {
         this(List.of(
-                requireThat("firstCard",  firstCard,  is(notNullValue())),
+                requireThat("firstCard", firstCard, is(notNullValue())),
                 requireThat("secondCard", secondCard, is(notNullValue()))
         ));
     }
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,127 @@
...
+package com.coderanch.blackjack;
+
+import java.util.*;
+
+/**
+ * A hand in a game of Blackjack.
+ */
+public class Hand {
+
+    private static final int HIGHEST_SCORE = 21;
```
nibsi: I'd name this constant `MAX_LEGAL_SCORE`.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final int MAX_SCORE_COUNT = 2^4;
```
nibsi: What does this constant mean? How did you come up with its value?

Please add documentation to all constants.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,127 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.coderanch.blackjack;
+
+import java.util.*;
+
+/**
+ * A hand in a game of Blackjack.
+ */
+public class Hand {
```
nibsi: Is this class used outside of its package yet? If not, remove the `public` specifier.

Has your class been designed for extensibility? If not, add the `final` modifier.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,127 @@
...
+    private boolean isStopped;
+    private Set<Integer> possibleScores;
+    private int bestScore;
+
+    /**
+     * Adds a card to the hand.
+     *
+     * @param card the card being added to the hand.
+     */
+    public void addCard(Card card) {
```
nibsi: I am not yet convinced that this class would be easier to use if it was mutable.

I suggest that you make this class immutable by removing all mutators. Add a method that creates a new hand when adding an extra card:

```java
Hand withAdditionalCard(Card card)
```

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -69,6 +77,10 @@ Suit suit() {
         return this.suit;
     }
 
+    int[] values() {
+        return this.rank.points;
```
nibsi: You are returning the private array of a card's rank. This means that any other class can do something like this:

```java
queenOfHearts.values()[0] = 5;
```
Suddenly, all queens are worth 5 points instead of 10.

In general, the following advice applies:

- Don't accept arrays in non-private methods and constructors. Accept collections instead.
- When accepting a collection in a non-private method or constructor, make a defensive copy.
- Never return arrays. Always return collections.
- (Almost) never return modifiable collections. Make collections unmodifiable before you return them.

In this particular case though, I don't think that using a collection is the easiest solution. I would give every rank one fixed value (give aces the value 11) and add special handling for aces in the `Hand` class.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,127 @@
...
+        if(isStopped){
```
nibsi: I would name a hand that has been passed on "fixed" rather than "stopped".

However, I don't think it should be the `Hand`'s responsibility to keep track of whether its fixed. It's the responsibility of the Game of Blackjack.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,69 @@
+package com.coderanch.blackjack;
```
nibsi, line 8: I'm missing a copyright notice.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public Hand split() {
+        if(!this.canSplit()) {
+            throw new IllegalHandSplitException("This hand cannot be split.");
+        }
+        var newHand = new Hand();
+        newHand.addCard(this.cards.remove(1));
+        this.refreshScore();
+        return newHand;
+    }
```
nibsi, line 93: Keep pull requests simple by using one PR per feature. I think we should defer splitting hands until after we can use our application to actually play a game of Blackjack using the most simple choices (hitting and passing only).

Revert the changes you've made that relate to splitting hands, but keep the code in a separate branch so we can always refer to it later.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        hand.addCard(new Card(Rank.KING, Suit.CLUBS));
+        assertThat(hand.bestScore(), is(equalTo(12)));
+
+        hand.addCard(new Card(Rank.EIGHT, Suit.CLUBS));
+        assertThat(hand.bestScore(), is(equalTo(20)));
+
+        hand.addCard(new Card(Rank.ACE, Suit.CLUBS));
+        assertThat(hand.bestScore(), is(equalTo(21)));
+    }
```
nibsi, line 277: The point of Theories is that you can test a general theory that applies to a wide range of data points. Currently you are testing one specific sequence of values.

I admit this is difficult if you haven't done it before, and it's hard for me to explain in a PR comment. Change the rest of your code first, then I will add a Theory that demonstrates what I mean.

nibsi (changes requested):

core/src/main/java/com/coderanch/blackjack/Hand.java
liutaurasvilda: @aquino-a are you aware `^` is a bitwise operator, so `2 ^ 4` would yield `6`. Is that expected?

aquino-a: oops, i wanted to do this: `Math.pow(2,4)`.   It's for the initial size of the `deque` used for calculating score.  Ace is the only card with 2 possible values.  I am assuming that there will be one deck per game which only has 4 aces.  The base score is calculated using the cards with one value, then the algorithm will double the amount of scores for every ace.

aquino-a: If i make ace a special case then the algorithm can be simplified and won't need this anymore

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -25,7 +25,15 @@
      * The rank of a card.
      */
     enum Rank {
-        ACE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE, TEN, JACK, QUEEN, KING
+        ACE(11), TWO(2), THREE(3), FOUR(4), FIVE(5),
+        SIX(6), SEVEN(7), EIGHT(8), NINE(9),
+        TEN(10), JACK(10), QUEEN(10), KING(10);
```
nibsi:
```suggestion
        ACE   (11),
        TWO   ( 2),
        THREE ( 3),
        FOUR  ( 4),
        FIVE  ( 5),
        SIX   ( 6),
        SEVEN ( 7),
        EIGHT ( 8),
        NINE  ( 9),
        TEN   (10),
        JACK  (10),
        QUEEN (10),
        KING  (10);
```

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -25,7 +25,15 @@
...
+
+        private int points;
```
nibsi:
```suggestion
        private final int points;
```

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Highest score in Blackjack.
+     */
+    private static final int MAX_LEGAL_SCORE = 21;
+
+    /**
+     * Max amount of aces
+     */
+    private static final int MAX_SCORE_COUNT = 4;
```
nibsi: The constant name is confusing, and you also don't need it because you can derive it from `Suit.values().length`.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+
+
+    private List<Card> cards;
+    private List<Integer> possibleScores;
+    private int bestScore;
```
nibsi: You don't need these two fields. Simply recalculate the value of a hand every time the `bestScore()` property is called.

If it turns out that calculating the score repeatedly causes performance issues, only then do we add fields that cache the calculated value. To determine the cause of performance issues you must use a profiler.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    private static final int MAX_LEGAL_SCORE = 21;
+
+    /**
+     * Max amount of aces
+     */
+    private static final int MAX_SCORE_COUNT = 4;
+
+
+    private List<Card> cards;
```
nibsi:
```suggestion
    private final List<Card> cards;
```

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+    private List<Integer> possibleScores;
+    private int bestScore;
+
+    /**
+     * Creates a new hand.
+     *
+     * @param card  first card dealt.
+     * @param card2 second card dealt
+     * @throws NullPointerException if either {@code card} or {@code card2} is {@code null}.
```
nibsi:
```suggestion
     * @param firstCard  the first card dealt.
     * @param secondCard the second card dealt.
     * @throws IllegalArgumentException if either {@code firstCard} or {@code secondCard} is {@code null}.
```

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public Hand(Card card, Card card2) {
+        this(List.of(card, card2));
```
nibsi:
```suggestion
    public Hand(Card firstCard, Card secondCard) {
        requireThat("firstCard",  firstCard,  is(notNullValue()));
        requireThat("secondCard", secondCard, is(notNullValue()));
        this(List.of(firstCard, secondCard));
```

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+                .forEach((Card c) -> {
+                    var initialSize = scores.size();
+                    for (int i = 0; i < initialSize; i++) {
+                        var score = scores.get(i);
+                        scores.set(i, score + 1);
+                        if (i + 1 == initialSize) {
+                            scores.add(score + c.points());
+                        }
+                    }
+                });
```
nibsi: Using `forEach` to fill a collection goes against the spirit of the streams API.

The preferred way to do it would be to write a custom collector. That can get pretty complex though, and you don't need to calculate a collection of all possible scores anyway, you only need the greatest score that is still legal.

Get rid of this method and simply let `calculateBestScore` figure out the the best score in a different way.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,108 @@
...
+        return scores;
+    }
+
+    private int calculateBestScore() {
```
nibsi: You can easily calculate the best score if you write a recursive function that looks like this:

```suggestion
    private static int calculateBestScore(Collection<Integer> fixedPointValues, int numberOfFreeAces) {
```
In the recursive step, convert free aces to fixed point values.

The base case is when you don't have free aces to convert to fixed point values.

If there are no legal scores, return the lowest score (having converted all aces to 1).

nibsi (changes requested):

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: I'm not against rearranging the ACE to appear below the KING, but I don't have a strong opinion about it.

core/src/main/java/com/coderanch/blackjack/Hand.java
aquino-a: I originally wanted to write it like this.  It won't compile because `this()` has to be called first.

core/src/main/java/com/coderanch/blackjack/Hand.java
aquino-a: I'm stumped on using fixed point values for this.  Maybe you can give me a hint.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -76,13 +96,16 @@ Suit suit() {
...
     }
 
+    int points() {
+        return this.rank.points;
+    }
+
     /**
      * Compares this card to another card.
-     *
+     * <p>
```
nibsi: I like that you're using paragraphs, but use them as intended:
```html
/**
 * <p>Compares this card to another card.</p>
 * <p>Cards are first compared by rank, then by suit.</p>

File: core/src/main/java/com/coderanch/blackjack/Hand.java [hunk 20]
```diff
@@ -0,0 +1,102 @@
...
+     * The cards in the hand.
+     */
+    private final List<Card> cards;
+
+    /**
+     * Creates a new hand.
+     *
+     * @param card  first card dealt.
+     * @param card2 second card dealt
+     * @throws NullPointerException if either {@code card} or {@code card2} is {@code null}.
```
nibsi: Throw `IllegalArgumentException` and use `requireThat` to validate the arguments.

aquino-a: It doesn't compile if you put `requireThat` before calling `this()`.  I auto commited the change you made but removed it.   Am I missing something?

aquino-a: It doesn't compile if you put requireThat before calling this(). I auto commited the change you made but removed it. Am I missing something?

core/src/main/java/com/coderanch/blackjack/Card.java
aquino-a: Actually I think the autoformat put that in there.  Do you want me to use `<p>`?

core/src/main/java/com/coderanch/blackjack/Hand.java
nibsi: Sure.

```java
    public int bestScore() {
        var minimumScore = cards.stream()
            .map(Card::rank)
            .filter(not(ACE::equals))
            .mapToInt(Rank::points)
            .sum();
        
        var numberOfFreeAces = (int) cards.stream()
            .map(Card::rank)
            .filter(ACE::equals)
            .count();
        
        return calculateBestScore(minimumScore, numberOfFreeAces);
    }

    private static int calculateBestScore(int minimumScore, int numberOfFreeAces) {
        // You still need to handle the base case here.
        
        var bestScoreWithBigAce = calculateBestScore(minimumScore + 11, numberOfFreeAces - 1);
        
        // You still need to do some more work here.
    }
```

core/src/main/java/com/coderanch/blackjack/Hand.java
nibsi:
```java
    /**
     * Creates a new hand.
     *
     * @param firstCard  the first card dealt.
     * @param secondCard the second card dealt
     * @throws IllegalArgumentException if either {@code firstCard} or {@code secondCard} is {@code null}.
     */
    Hand(Card firstCard, Card secondCard) {
        this(List.of(
            requireThat("firstCard",  firstCard,  is(notNullValue())),
            requireThat("secondCard", secondCard, is(notNullValue()))
        ));
    }
```

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: What do you mean by autoformat? Is it an IDE feature?

I want you to use `<p>` if it makes the generated Javadoc look more readable, but I want you to use it like proper semantic HTML, meaning the summary consists of two paragraphs, NOT two loose pieces of text separated by an empty paragraph.

If your IDE is not capable of producing semantic HTML for Javadoc comments, then don't use it to format Javadoc.

When in doubt, just remove all HTML from your Javadoc, and we'll fix it when we focus on making the generated Javadoc look good.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -25,7 +25,32 @@
...
+        /**
+         * The points value of the rank.
+         */
+        private final int points;
+
+        Rank(int points) {
+            this.points = points;
+        }
+
+        public int getPoints() {
```
nibsi: Inconsistent naming in regards to `Card.points()`.

Name this method `points` as well.

nibsi: Why did you add a public method to a non-public type?

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -50,7 +75,6 @@
      *
      * @param rank the rank of the new card.
      * @param suit the suit of the new card.
-     *
```
nibsi: Why all these changes?

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.coderanch.blackjack;
+
+import java.util.*;
+
+import static com.coderanch.blackjack.Card.Rank.ACE;
+import static com.coderanch.util.require.Require.requireThat;
+import static java.util.function.Predicate.not;
```
nibsi: Sort and group your import statements. Put imports from java packages at the top, regardless of alphabetical order.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+     * Creates a new hand.
+     *
+     * @param firstCard  first card dealt.
+     * @param secondCard second card dealt
+     * @throws IllegalArgumentException if either {@code firstCard} or {@code secondCard} is {@code null}.
+     */
+    Hand(Card firstCard, Card secondCard) {
+        this(List.of(
+                requireThat("firstCard", firstCard, is(notNullValue())),
+                requireThat("secondCard", secondCard, is(notNullValue()))
```
nibsi: Set your continuation indentation to 4 spaces, not 8.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+        ));
+    }
+
+    private Hand(List<Card> cards) {
```
nibsi, line 55: Missing documentation.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+        var minimumScore = cards.stream()
+                .map(Card::rank)
+                .filter(not(ACE::equals))
+                .mapToInt(Card.Rank::getPoints)
+                .sum();
+
+        var numberOfFreeAces = (int) cards.stream()
+                .map(Card::rank)
+                .filter(ACE::equals)
+                .count();
```
nibsi: More indentation woes.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+
+        var bestScore = calculateBestScore(minimumScore, numberOfFreeAces);
+        if (bestScore > MAX_LEGAL_SCORE) {
+            return 0;
```
nibsi: Why 0? What's wrong with just returning the actual score?

Add a method `isBust()` that returns `true` when `bestScore()` is greater than `MAX_LEGAL_SCORE`.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+        else {
+            return bestScore;
+        }
+    }
+
+    private static int calculateBestScore(int minimumScore, int numberOfFreeAces) {
+        if (numberOfFreeAces <= 0) {
```
nibsi: In what case would `numberOfFreeAces` ever be less than 0?

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+            return minimumScore;
+        }
+        var bestScoreWithBigAce = calculateBestScore(minimumScore + ACE.getPoints(), numberOfFreeAces - 1);
+
+        if (bestScoreWithBigAce > MAX_LEGAL_SCORE) {
+            return calculateBestScore(minimumScore + 1, numberOfFreeAces - 1);
+        }
+        else {
+            return bestScoreWithBigAce;
+        }
```
nibsi:
```suggestion

        var bestScoreWithBigAce = calculateBestScore(minimumScore + ACE.getPoints(), numberOfFreeAces - 1);

        return bestScoreWithBigAce > MAX_LEGAL_SCORE
            ? calculateBestScore(minimumScore + 1, numberOfFreeAces - 1);
            : bestScoreWithBigAce;
```

nibsi: Good implementation!

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+@RunWith(Theories.class)
+public final class HandTest {
+
+    /**
+     * Hands to test scoring.
+     */
+    @DataPoints
+    public static final List<HandTestArgument> HANDS =
+            List.of(
```
nibsi: Fix the indentation in this class. 4 spaces per level is enough.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Gets a new hand with the extra card.
+     *
+     * @param card the card being added to the hand.
+     * @return a new hand with the extra card.
+     * @throws IllegalArgumentException if {@code card} is {@code null}.
+     */
+    public Hand withAdditionalCard(Card card) {
```
nibsi: More public methods in non-public class.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+                                    new Card(Rank.NINE, Suit.CLUBS)),
+                            21
+                    )
+            );
+
+    /**
+     * Cards to test.
+     */
+    @DataPoints("objects")
+    public static final Set<Card> CARDS = Cards.getStandardDeck();
```
nibsi, line 174: The `"objects"` data point is meant to provide instances of the class under test. For `HandTest`, that would be a set of hands, not a set of cards.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+
+    /**
+     * Tests that {@link Hand#withAdditionalCard(Card)} maintains the correct score.
+     *
+     * @param handTest
```
nibsi: Missing parameter description.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @SuppressWarnings("checkstyle:methodname")
+    public void addCard_maintainsCorrectScore(HandTestArgument handTest) {
+        var hand = new Hand(handTest.getCards().get(0), handTest.getCards().get(1));
+        for (int i = 2; i < handTest.getCards().size(); i++) {
+            hand = hand.withAdditionalCard(handTest.getCards().get(i));
+        }
+        assertThat("The best score must match the target", hand.bestScore(), is(handTest.getTargetScore()));
+    }
+
+    private static final class HandTestArgument {
```
nibsi: Don't mix classes and methods. Either put this class at the top, or put it at the bottom.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Tests that passing {@code null} for {@code card} when constructing a new hand causes an exception to be thrown.
+     *
+     * @param card the card to construct the hand with.
+     */
+    @Theory(nullsAccepted = false)
+    @SuppressWarnings("checkstyle:methodname")
+    public void newHand_withNullCard1_throwsException(Card card) {
```
nibsi: `newHand_withNullFirstCard_throwsException`

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+            return cards;
+        }
+
+        int getTargetScore() {
+            return targetScore;
+        }
+    }
+
+    /**
+     * Tests that passing {@code null} for {@code card} when constructing a new hand causes an exception to be thrown.
```
nibsi: Misleading documentation. The assertion is not about `card`, but rather about the first card of the hand.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @Theory(nullsAccepted = false)
+    @SuppressWarnings("checkstyle:methodname")
+    public void newHand_withNullCard1_throwsException(Card card) {
+        assertThrows("Hand must throw exception", IllegalArgumentException.class, () -> {
+            new Hand(null, card);
+        });
+    }
+
+    /**
+     * Tests that passing {@code null} for {@code card} when constructing a new hand causes an exception to be thrown.
```
nibsi: Same comments as previous method.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @Theory
+    @SuppressWarnings("checkstyle:methodname")
+    public void addCard_withNullCard_throwsException() {
+        var hand = new Hand(new Card(Rank.ACE, Suit.CLUBS), new Card(Rank.ACE, Suit.CLUBS));
+        assertThrows("Hand must throw exception", IllegalArgumentException.class, () -> {
+            hand.withAdditionalCard(null);
+        });
+    }
+
+
```
nibsi, line 294: Spurious newline.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    Hand(Card firstCard, Card secondCard) {
+        this(List.of(
+                requireThat("firstCard", firstCard, is(notNullValue())),
+                requireThat("secondCard", secondCard, is(notNullValue()))
+        ));
+    }
+
+    private Hand(List<Card> cards) {
+        this.cards = cards;
```
nibsi, line 56: This is fine. Ignore Sonar. I will change the rule.

nibsi (changes requested):

core/src/main/java/com/coderanch/blackjack/Card.java
aquino-a: These kinds of changes were made when auto formatting.

core/src/main/java/com/coderanch/blackjack/Card.java
aquino-a: I'm not used to using package private so I did it out of bad habit.

core/src/main/java/com/coderanch/blackjack/Hand.java
aquino-a: haha I guess it won't ever be. just in case

core/src/main/java/com/coderanch/blackjack/Hand.java
aquino-a: thanks

core/src/test/java/com/coderanch/blackjack/HandTest.java
aquino-a: I see. So I should remove it or refactor the tests?

nibsi: I admit I still have to look over the tests a bit more, but you probably don't have to refactor it for now. Just remove the `"objects"` identifier from the data point.

core/src/main/java/com/coderanch/blackjack/Card.java
nibsi: Please revert them. Auto-formatting should only affect lines you've done work on, not lines that are unrelated to your work item.

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -76,13 +100,15 @@ Suit suit() {
         return this.suit;
     }
 
+    int points() {
+        return this.rank.points;
+    }
```
nibsi: Seeing as this method isn't covered by code coverage, I'm going to assume that it's not used. Please remove it.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * A hand in a game of Blackjack.
+ */
+final class Hand {
+
+    /**
+     * Highest score in Blackjack.
+     */
+    static final int MAX_LEGAL_SCORE = 21;
```
nibsi: Make this constant private.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,113 @@
...
+
+        var newCards = new ArrayList<Card>(this.cards);
+        newCards.add(card);
+        return new Hand(newCards);
+    }
+
+    /**
+     * Gets the best score.
+     *
+     * @return the best score.
```
nibsi: Requires more explanation.

```suggestion
     * @return the highest score that can be made by using a large point value for aces
     *         without the hand going bust, or the lowest score if the hand is bust.
```

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,113 @@
...
+
+        return bestScoreWithBigAce > MAX_LEGAL_SCORE
+            ? calculateBestScore(minimumScore + 1, numberOfFreeAces - 1)
+            : bestScoreWithBigAce;
+    }
+
+    /**
+     * Get whether the hand is bust or not.
+     *
+     * @return whether the hand is bust or not.
```
nibsi:
```suggestion
     * Gets whether the hand is bust or not.
     * A bust hand has a higher score than the legal maximum.
     *
     * @return {@code true} if {@link #bestScore()} is greater than the maximum legal score;
     *         {@code false} otherwise.
```

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.Set;
+
+import com.coderanch.blackjack.Card.Rank;
+import com.coderanch.blackjack.Card.Suit;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+import static org.junit.Assert.assertThrows;
```
nibsi: Put static imports below non-static imports.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+
+    /**
+     * Tests that {@link Hand#withAdditionalCard(Card)} maintains the correct score.
+     *
+     * @param handTest contains the test hand, and the target score.
+     */
+    @Theory
+    @SuppressWarnings("checkstyle:methodname")
+    public void addCard_maintainsCorrectScore(HandTestArgument handTest) {
```
nibsi: A test should use the method it tests directly.

Inject a hand and a card as method parameters. Assume that the hand contains no aces, and that the card to add is not an ace.

Call `withAdditionalCard()`.

Test that the resulting hand has the expected difference in points when compared to the original hand.

Also, there is no `addCard` method, so rename this test.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,200 @@
...
+        assertThat(
+            "The best score must match the target",
+            handTest.getHand().bestScore(),
+            is(handTest.getTargetScore())
+        );
+    }
+
+    /**
+     * Tests that passing {@code null} for the first {@code card}
```
nibsi: Why is `card` in a code tag? Just write prose.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,200 @@
...
+     * Tests that hands that {@link Hand#isBust()} returns the correct answer.
+     *
+     * @param handTest contains the test hand, and the target score.
+     */
+    @Theory
+    public void isBust(HandTestArgument handTest) {
+        assertThat(
+            "The hands over the legal limit must be bust.",
+            handTest.getHand().isBust(),
+            is(equalTo(handTest.getHand().bestScore() > Hand.MAX_LEGAL_SCORE))
```
nibsi: I'd prefer it that you create a data point that consists of various hands that you know are bust, and then test that for these hands `isBust()` returns true.

Add another test that does the same thing with hands that you know ahead of time are not bust.

nibsi (changes requested):
Almost there!

core/src/main/java/com/coderanch/blackjack/Card.java
```diff
@@ -25,7 +25,32 @@
...
+         */
+        private final int points;
+
+        Rank(int points) {
+            this.points = points;
+        }
+
+        int points() {
+            return points;
+        }
```
nibsi: Missing documentation on these two members.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,349 @@
...
+            hand = hand.withAdditionalCard(cards.get(i));
+        }
+        return hand;
+    }
+
+    /**
+     * Cards to test.
+     */
+    @DataPoints
+    public static final Set<Card> CARDS = Cards.getStandardDeck();
```
nibsi: Don't mix fields and methods.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,349 @@
...
+        });
+    }
+
+    /**
+     * Tests that hands that {@link Hand#isBust()} returns the correct answer.
+     *
+     * @param bustHand the hand to check if bust.
+     */
+    @Theory
+    public void isBust(BustHand bustHand) {
```
nibsi: Instead of creating separate classes for `BustHand` and `NonBustHand`, just use the `Hand` class. You can indicate from which data points of type `List<Hand>` you want to select your test arguments by giving your data points names and then using [`@FromDataPoints`](https://junit.org/junit4/javadoc/latest/org/junit/experimental/theories/FromDataPoints.html) in your theories.

nibsi: `isBust_withBustHand_returnsTrue`

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,349 @@
...
+        assertThat(
+            "The hand must be bust.",
+            bustHand.hand.isBust(),
+            is(equalTo(true))
```
nibsi: `equalTo` is mostly used to make the matcher read more fluently. It doesn't actually do anything. So in cases like this, you can just write `is(true)`.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,349 @@
...
+        );
+    }
+
+    /**
+     * Tests that hands that {@link Hand#isBust()} returns the correct answer.
+     *
+     * @param nonBustHand the hand to check if not bust.
+     */
+    @Theory
+    public void isNotBust(NonBustHand nonBustHand) {
```
nibsi: `isBust_withNonBustHand_returnsFalse`

nibsi (changes requested):
Very nice. I think with your next update you'll be done.

Don't forget to merge the `master` branch into this branch before you request your next review, otherwise you won't be able to finish the pull request.

core/src/test/java/com/coderanch/blackjack/HandTest.java
aquino-a: I thought there must have been an easier way but didn't look hard enough I guess.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -209,7 +209,7 @@ public void bestScore_hasCorrectScore(BestScoreTestArgument scoreTest) {
      */
     @Theory
     @SuppressWarnings("checkstyle:methodname")
-    public void withAdditionalCard_hasCorrectScore(Hand hand, Card card) {
+    public void withAdditionalCard_hasCorrectScore(@FromDataPoints("Add card hands") Hand hand, Card card) {
         if (card.rank() == Rank.ACE) {
             return;
         }
```
nibsi: Sorry, I should have caught this earlier.

When testing with Theories, guard against arguments you're not interested in using an assumption, not an if-statement:

```suggestion
        assumeThat(card.rank(), is(not(Rank.ACE)));
```

The practical difference between the two is that when you pass an Ace, an if-statement will report the test as having succeeded, while the assumption will report the test as having been skipped.

nibsi (dismissed):

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,324 @@
...
+import org.junit.experimental.theories.DataPoints;
+import org.junit.experimental.theories.FromDataPoints;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
+import org.junit.runner.RunWith;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.equalTo;
```
nibsi: Put this import before the import to `is`.

core/src/test/java/com/coderanch/blackjack/HandTest.java
```diff
@@ -0,0 +1,324 @@
...
+
+import static org.junit.Assume.assumeThat;
+import static org.junit.Assert.assertThrows;
```
nibsi: Swap these two imports.

nibsi (dismissed):

core/src/test/java/com/coderanch/blackjack/HandTest.java
aquino-a: not sure what you mean

nibsi: I mean that within an import group, the imports should be sorted alphabetically, so put `equalTo` before `is`.

nibsi (dismissed):

nibsi (approved):

nibsi:
@liutaurasvilda can you give this one more quick glance? I think we can finish it.

PR #20, opened 2021-03-17
add input utility

### LINKED ISSUE #17: Create a multipurpose input utility for the CLI
# Multipurpose Input Utility

Create a multipurpose input utility for use with a command line interface.  The user can pass a prompt and optional validation and the utility will read the provided input stream and return a possibly validated String or primitive.

### InputUtility([input_stream])
Constructor that takes an input stream to be used.  If no input stream is provided, System.in will be used.

### nextString(prompt, [validation_predicate])
Displays the prompt and reads the input stream, returning a possibly validated String.

### nextInt(prompt, [validation_int_predicate])
Displays the prompt and reads the input stream, returning a possibly validated integer.

### nextDouble(prompt, [validation_double_predicate])
Displays the prompt and reads the input stream, returning a possibly validated real number.

### nextYesNo(prompt, [validation_yes_no_predicate])
Calls nextString, passing the prompt and a predicate that validates (case insensitive) "y", "n", "yes", "no".

### pause([prompt])
Calls nextString with the prompt entered and waits for the user to press <enter>.  The prompt defaults to "Press <enter> when ready" if no prompt is passed.

### yesOrNo()
Returns a String predicate that tests for (case insensitive) "y", "n", "yes", "no".

### oneOfThese(String...)
Returns a String predicate that takes two or more Strings and tests whether any one of them match the input.

### intRange(lower, upper)
Returns an int predicate that tests whether the input is between the two limits.

### doubleRange(lower, upper)
Returns a double predicate that tests whether the input is between the two limits.

#

### LINKED ISSUE #33: Change/enhance "yes or no" processing in InputUtility
Currently, when the `InputUtility` method `getYesNo` is used, it forces the user to enter a "yes or no" predicate and hope that the user's predicate has a "yes" value that matches the `YES` list constant. Instead, the utility should use the internal `yesOrNo` predicate to make sure only a positive or negative response was inputted.

aquino-a:
resolves #17

### DIFF
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -1,6 +1,5 @@
 package com.coderanch.util.cli;
 
-import junit.framework.TestCase;
 import org.junit.experimental.theories.Theories;
 import org.junit.experimental.theories.Theory;
 import org.junit.runner.RunWith;
@@ -11,6 +10,8 @@ import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Predicate;
 
 /**
  * Tests the {@link InputUtility} class.
@@ -20,84 +21,74 @@ public class InputUtilityTest {
 
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#nextString(String, Predicate)} returns the expected value.
      */
     @Theory()
     public void nextString_validatesCorrectInput() throws IOException {
-        var testCase = "input";
-        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s);
-            var result = iu.nextString("prompt", s1 -> s1.equals(testCase));
-            assertThat("Matches testCase", testCase, is(equalTo(result)));
-        }
+        withInput("input", inputUtility -> {
+            var string = inputUtility.nextString("prompt", s -> true);
+            assertThat("Next string must match expected value.", string, is("input"));
+        });
     }
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#nextInt(String, Predicate)} returns the expected value.
      */
     @Theory()
     public void nextInt_validatesCorrectInput() throws IOException {
-        var testCase = "1";
-        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s);
-            var result = iu.nextInt("prompt", n -> n == 1);
-            assertThat("Matches testCase", Integer.parseInt(testCase), is(equalTo(result)));
-        }
+        withInput("1", inputUtility -> {
+            var integer = inputUtility.nextInt("prompt", i -> true);
+            assertThat("Next int must match expected value.", integer, is(1));
+        });
     }
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#nextDouble(String, Predicate)} returns the expected value.
      */
     @Theory()
     public void nextDouble_validatesCorrectInput() throws IOException {
-        var testCase = "2.2";
-        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s);
-            var result = iu.nextDouble("prompt", n -> n == 2.2d);
-            assertThat("Matches testCase", Double.parseDouble(testCase), is(equalTo(result)));
-        }
+        withInput("2.2", inputUtility -> {
+            var dble = inputUtility.nextDouble("prompt", d -> true);
+            assertThat("Next int must match expected value.", dble, is(2.2));
+        });
     }
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#nextYesNo(String, Predicate)} returns the expected value.
      */
     @Theory()
     public void nextYesNo_validatesYesCorrectly() throws IOException {
-        var testCase = " yEs ";
-        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s);
-            var result = iu.nextYesNo("prompt", InputUtility.yesOrNo());
-            assertThat("Matches testCase", true, is(equalTo(result)));
-        }
+        withInput(" yEs ", inputUtility -> {
+            var bool = inputUtility.nextYesNo("prompt", InputUtility.yesOrNo());
+            assertThat("Next yesNo must match expected value.", bool, is(true));
+        });
     }
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#nextYesNo(String, Predicate)} returns the expected value.
      */
     @Theory()
     public void nextYesNo_validatesNoCorrectly() throws IOException {
-        var testCase = "NO";
-        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s);
-            var result = iu.nextYesNo("prompt", InputUtility.yesOrNo());
-            assertThat("Matches testCase", false, is(equalTo(result)));
-        }
+        withInput("NO", inputUtility -> {
+            var bool = inputUtility.nextYesNo("prompt", InputUtility.yesOrNo());
+            assertThat("Next yesNo must match expected value.", bool, is(false));
+        });
     }
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#pause()} works.
      */
     @Theory()
     public void pause_validatesCorrectly() throws IOException {
         var testCase = ".";
         try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s);
+            var iu = new InputUtility(s, StandardCharsets.UTF_8);
             iu.pause();
         }
     }
 
     /**
-     * @throws IOException
+     * Tests that {@link InputUtility#yesOrNo()} returns the expected value.
      */
     @Theory()
     public void yesOrNo_validatesCorrectly() throws IOException {
@@ -114,13 +105,42 @@ public class InputUtilityTest {
         assertThat("Matches correctly", true, is(equalTo(result4)));
     }
 
+    /**
+     * Tests that {@link InputUtility#oneOfThese(String...)} returns the expected value.
+     */
+    @Theory()
+    public void oneOfThese_validatesCorrectly() throws IOException {
 
-    public void testOneOfThese() {
     }
 
-    public void testIntRange() {
+    /**
+     * Tests that {@link InputUtility#intRange(int, int)} returns the expected value.
+     */
+    @Theory()
+    public void intRange_validatesCorrectly() throws IOException {
     }
 
-    public void testDoubleRange() {
+    /**
+     * Tests that {@link InputUtility#doubleRange(double, double)} returns the expected value.
+     */
+    @Theory()
+    public void doubleRange_validatesCorrectly() throws IOException {
+    }
+
+    private static <X extends Throwable> void withInput(
+            String input,
+            ExceptionalConsumer<? super InputUtility, X> action
+    ) throws IOException, X {
+        try (var stream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8))) {
+            var inputUtility = new InputUtility(stream, StandardCharsets.UTF_8);
+            action.accept(inputUtility);
+        }
+    }
+
+    @FunctionalInterface
+    private static interface ExceptionalConsumer<T, X extends Throwable> {
+
+        void accept(T argument) throws X;
+
     }
 }
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -127,6 +127,7 @@ public class InputUtilityTest {
 
     /**
      * Tests that {@link InputUtility#yesOrNo()} returns the expected value.
+     *
      * @param input user input to check.
      */
     @Theory
@@ -138,6 +139,7 @@ public class InputUtilityTest {
 
     /**
      * Tests that {@link InputUtility#oneOfThese(String...)} returns the expected value.
+     *
      * @param values the strings to check against in the predicate.
      */
     @Theory
@@ -149,6 +151,7 @@ public class InputUtilityTest {
 
     /**
      * Tests that {@link InputUtility#intRange(int, int)} returns the expected value.
+     *
      * @param values the number range check against in the predicate.
      */
     @Theory
@@ -160,6 +163,7 @@ public class InputUtilityTest {
 
     /**
      * Tests that {@link InputUtility#doubleRange(double, double)} returns the expected value.
+     *
      * @param values the number range check against in the predicate.
      */
     @Theory
@@ -173,8 +177,10 @@ public class InputUtilityTest {
             String input,
             ExceptionalConsumer<? super InputUtility, X> action
     ) throws IOException, X {
-        try (var stream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8))) {
-            var inputUtility = new InputUtility(stream, StandardCharsets.UTF_8);
+        try (
+                var stream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8));
+                var inputUtility = new InputUtility(stream, StandardCharsets.UTF_8);
+        ) {
             action.accept(inputUtility);
         }
     }
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -23,7 +23,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#nextString(String, Predicate)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void nextString_validatesCorrectInput() throws IOException {
         withInput("input", inputUtility -> {
             var string = inputUtility.nextString("prompt", s -> true);
@@ -34,7 +34,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#nextInt(String, Predicate)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void nextInt_validatesCorrectInput() throws IOException {
         withInput("1", inputUtility -> {
             var integer = inputUtility.nextInt("prompt", i -> true);
@@ -45,7 +45,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#nextDouble(String, Predicate)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void nextDouble_validatesCorrectInput() throws IOException {
         withInput("2.2", inputUtility -> {
             var dble = inputUtility.nextDouble("prompt", d -> true);
@@ -56,7 +56,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#nextYesNo(String, Predicate)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void nextYesNo_validatesYesCorrectly() throws IOException {
         withInput(" yEs ", inputUtility -> {
             var bool = inputUtility.nextYesNo("prompt", InputUtility.yesOrNo());
@@ -67,7 +67,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#nextYesNo(String, Predicate)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void nextYesNo_validatesNoCorrectly() throws IOException {
         withInput("NO", inputUtility -> {
             var bool = inputUtility.nextYesNo("prompt", InputUtility.yesOrNo());
@@ -78,7 +78,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#pause()} works.
      */
-    @Theory()
+    @Theory
     public void pause_validatesCorrectly() throws IOException {
         var testCase = ".";
         try (var s = new ByteArrayInputStream(testCase.getBytes())) {
@@ -90,7 +90,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#yesOrNo()} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void yesOrNo_validatesCorrectly() throws IOException {
         var result = InputUtility.yesOrNo().test("   YEs ");
         assertThat("Matches correctly", true, is(equalTo(result)));
@@ -108,7 +108,7 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#oneOfThese(String...)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void oneOfThese_validatesCorrectly() throws IOException {
 
     }
@@ -116,14 +116,14 @@ public class InputUtilityTest {
     /**
      * Tests that {@link InputUtility#intRange(int, int)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void intRange_validatesCorrectly() throws IOException {
     }
 
     /**
      * Tests that {@link InputUtility#doubleRange(double, double)} returns the expected value.
      */
-    @Theory()
+    @Theory
     public void doubleRange_validatesCorrectly() throws IOException {
     }
 
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -1,16 +1,24 @@
+/*
+ * Copyright (C) 2018 Coderanch.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
 package com.coderanch.util.cli;
 
+import org.junit.experimental.theories.DataPoints;
 import org.junit.experimental.theories.Theories;
 import org.junit.experimental.theories.Theory;
 import org.junit.runner.RunWith;
 
-import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 import java.util.function.Predicate;
 
 /**
@@ -19,6 +27,32 @@ import java.util.function.Predicate;
 @RunWith(Theories.class)
 public class InputUtilityTest {
 
+    /**
+     * Inputs to test for yes and no.
+     */
+    @DataPoints
+    public static final List<String> YES_NO = List.of("   YEs ", "no", " n", "  y ");
+
+    /**
+     * Inputs to test for one of these.
+     */
+    @DataPoints
+    public static final List<List<String>> ONE_OF_THESE
+            = List.of(List.of("in1", "in2", "in3"), List.of("in1a", "in2a", "in3a"));
+
+    /**
+     * Inputs to test for intRange
+     */
+    @DataPoints
+    public static final int[][] INT_RANGE
+            = new int[][]{{1, 5}, {5, 7}};
+
+    /**
+     * Inputs to test for doubleRange
+     */
+    @DataPoints
+    public static final double[][] DOUBLE_RANGE
+            = new double[][]{{1.2, 5.2}, {2.3, 6.7}};
 
     /**
      * Tests that {@link InputUtility#nextString(String, Predicate)} returns the expected value.
@@ -80,51 +114,45 @@ public class InputUtilityTest {
      */
     @Theory
     public void pause_validatesCorrectly() throws IOException {
-        var testCase = ".";
-        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
-            var iu = new InputUtility(s, StandardCharsets.UTF_8);
-            iu.pause();
-        }
+        withInput(" ", inputUtility -> {
+            inputUtility.pause();
+        });
     }
 
     /**
      * Tests that {@link InputUtility#yesOrNo()} returns the expected value.
      */
     @Theory
-    public void yesOrNo_validatesCorrectly() throws IOException {
-        var result = InputUtility.yesOrNo().test("   YEs ");
-        assertThat("Matches correctly", true, is(equalTo(result)));
-
-        var result2 = InputUtility.yesOrNo().test("no");
-        assertThat("Matches correctly", true, is(equalTo(result2)));
-
-        var result3 = InputUtility.yesOrNo().test(" n");
-        assertThat("Matches correctly", true, is(equalTo(result3)));
-
-        var result4 = InputUtility.yesOrNo().test("  y ");
-        assertThat("Matches correctly", true, is(equalTo(result4)));
+    public void yesOrNo_validatesCorrectly(String input) {
+        var result = InputUtility.yesOrNo().test(input);
+        assertThat("Predicate must match expected value.", result, is(true));
     }
 
     /**
      * Tests that {@link InputUtility#oneOfThese(String...)} returns the expected value.
      */
     @Theory
-    public void oneOfThese_validatesCorrectly() throws IOException {
-
+    public void oneOfThese_validatesCorrectly(List<String> values) {
+        var result = InputUtility.oneOfThese(values.toArray(String[]::new)).test(values.get(0));
+        assertThat("Predicate must match expected value.", result, is(true));
     }
 
     /**
      * Tests that {@link InputUtility#intRange(int, int)} returns the expected value.
      */
     @Theory
-    public void intRange_validatesCorrectly() throws IOException {
+    public void intRange_validatesCorrectly(int[] values) throws IOException {
+        var result = InputUtility.intRange(values[0], values[1]).test(values[0] + 1);
+        assertThat("Predicate must match expected value.", result, is(true));
     }
 
     /**
      * Tests that {@link InputUtility#doubleRange(double, double)} returns the expected value.
      */
     @Theory
-    public void doubleRange_validatesCorrectly() throws IOException {
+    public void doubleRange_validatesCorrectly(double[] values) throws IOException {
+        var result = InputUtility.doubleRange(values[0], values[1]).test(values[0] + 1);
+        assertThat("Predicate must match expected value.", result, is(true));
     }
 
     private static <X extends Throwable> void withInput(
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -25,9 +25,9 @@ public class InputUtilityTest {
     @Theory()
     public void nextString_validatesCorrectInput() throws IOException {
         var testCase = "input";
-        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
             var iu = new InputUtility(s);
-            var result =  iu.nextString("prompt", s1 -> s1.equals(testCase));
+            var result = iu.nextString("prompt", s1 -> s1.equals(testCase));
             assertThat("Matches testCase", testCase, is(equalTo(result)));
         }
     }
@@ -38,9 +38,9 @@ public class InputUtilityTest {
     @Theory()
     public void nextInt_validatesCorrectInput() throws IOException {
         var testCase = "1";
-        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
             var iu = new InputUtility(s);
-            var result =  iu.nextInt("prompt", n -> n == 1);
+            var result = iu.nextInt("prompt", n -> n == 1);
             assertThat("Matches testCase", Integer.parseInt(testCase), is(equalTo(result)));
         }
     }
@@ -51,9 +51,9 @@ public class InputUtilityTest {
     @Theory()
     public void nextDouble_validatesCorrectInput() throws IOException {
         var testCase = "2.2";
-        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
             var iu = new InputUtility(s);
-            var result =  iu.nextDouble("prompt", n -> n == 2.2d);
+            var result = iu.nextDouble("prompt", n -> n == 2.2d);
             assertThat("Matches testCase", Double.parseDouble(testCase), is(equalTo(result)));
         }
     }
@@ -64,9 +64,9 @@ public class InputUtilityTest {
     @Theory()
     public void nextYesNo_validatesYesCorrectly() throws IOException {
         var testCase = " yEs ";
-        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
             var iu = new InputUtility(s);
-            var result =  iu.nextYesNo("prompt", InputUtility.yesOrNo());
+            var result = iu.nextYesNo("prompt", InputUtility.yesOrNo());
             assertThat("Matches testCase", true, is(equalTo(result)));
         }
     }
@@ -77,9 +77,9 @@ public class InputUtilityTest {
     @Theory()
     public void nextYesNo_validatesNoCorrectly() throws IOException {
         var testCase = "NO";
-        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
             var iu = new InputUtility(s);
-            var result =  iu.nextYesNo("prompt", InputUtility.yesOrNo());
+            var result = iu.nextYesNo("prompt", InputUtility.yesOrNo());
             assertThat("Matches testCase", false, is(equalTo(result)));
         }
     }
@@ -90,7 +90,7 @@ public class InputUtilityTest {
     @Theory()
     public void pause_validatesCorrectly() throws IOException {
         var testCase = ".";
-        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+        try (var s = new ByteArrayInputStream(testCase.getBytes())) {
             var iu = new InputUtility(s);
             iu.pause();
         }
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -1,0 +1,126 @@
+package com.coderanch.util.cli;
+
+import junit.framework.TestCase;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
+import org.junit.runner.RunWith;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+
+/**
+ * Tests the {@link InputUtility} class.
+ */
+@RunWith(Theories.class)
+public class InputUtilityTest {
+
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void nextString_validatesCorrectInput() throws IOException {
+        var testCase = "input";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            var result =  iu.nextString("prompt", s1 -> s1.equals(testCase));
+            assertThat("Matches testCase", testCase, is(equalTo(result)));
+        }
+    }
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void nextInt_validatesCorrectInput() throws IOException {
+        var testCase = "1";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            var result =  iu.nextInt("prompt", n -> n == 1);
+            assertThat("Matches testCase", Integer.parseInt(testCase), is(equalTo(result)));
+        }
+    }
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void nextDouble_validatesCorrectInput() throws IOException {
+        var testCase = "2.2";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            var result =  iu.nextDouble("prompt", n -> n == 2.2d);
+            assertThat("Matches testCase", Double.parseDouble(testCase), is(equalTo(result)));
+        }
+    }
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void nextYesNo_validatesYesCorrectly() throws IOException {
+        var testCase = " yEs ";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            var result =  iu.nextYesNo("prompt", InputUtility.yesOrNo());
+            assertThat("Matches testCase", true, is(equalTo(result)));
+        }
+    }
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void nextYesNo_validatesNoCorrectly() throws IOException {
+        var testCase = "NO";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            var result =  iu.nextYesNo("prompt", InputUtility.yesOrNo());
+            assertThat("Matches testCase", false, is(equalTo(result)));
+        }
+    }
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void pause_validatesCorrectly() throws IOException {
+        var testCase = ".";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            iu.pause();
+        }
+    }
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void yesOrNo_validatesCorrectly() throws IOException {
+        var result = InputUtility.yesOrNo().test("   YEs ");
+        assertThat("Matches correctly", true, is(equalTo(result)));
+
+        var result2 = InputUtility.yesOrNo().test("no");
+        assertThat("Matches correctly", true, is(equalTo(result2)));
+
+        var result3 = InputUtility.yesOrNo().test(" n");
+        assertThat("Matches correctly", true, is(equalTo(result3)));
+
+        var result4 = InputUtility.yesOrNo().test("  y ");
+        assertThat("Matches correctly", true, is(equalTo(result4)));
+    }
+
+
+    public void testOneOfThese() {
+    }
+
+    public void testIntRange() {
+    }
+
+    public void testDoubleRange() {
+    }
+}
--- a/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
+++ b/core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
@@ -19,7 +19,6 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
-import java.util.function.Predicate;
 
 /**
  * Tests the {@link InputUtility} class.
@@ -41,14 +40,14 @@ public class InputUtilityTest {
             = List.of(List.of("in1", "in2", "in3"), List.of("in1a", "in2a", "in3a"));
 
     /**
-     * Inputs to test for intRange
+     * Inputs to test for intRange.
      */
     @DataPoints
     public static final int[][] INT_RANGE
             = new int[][]{{1, 5}, {5, 7}};
 
     /**
-     * Inputs to test for doubleRange
+     * Inputs to test for doubleRange.
      */
     @DataPoints
     public static final double[][] DOUBLE_RANGE
@@ -58,6 +57,7 @@ public class InputUtilityTest {
      * Tests that {@link InputUtility#nextString(String, Predicate)} returns the expected value.
      */
     @Theory
+    @SuppressWarnings("checkstyle:methodname")
     public void nextString_validatesCorrectInput() throws IOException {
         withInput("input", inputUtility -> {
             var string = inputUtility.nextString("prompt", s -> true);
@@ -69,6 +69,7 @@ public class InputUtilityTest {
      * Tests that {@link InputUtility#nextInt(String, Predicate)} returns the expected value.
      */
     @Theory
+    @SuppressWarnings("checkstyle:methodname")
     public void nextInt_validatesCorrectInput() throws IOException {
         withInput("1", inputUtility -> {
             var integer = inputUtility.nextInt("prompt", i -> true);
@@ -80,6 +81,7 @@ public class InputUtilityTest {
      * Tests that {@link InputUtility#nextDouble(String, Predicate)} returns the expected value.
      */
     @Theory
+    @SuppressWarnings({"checkstyle:methodname", "checkstyle:magicnumber"})
     public void nextDouble_validatesCorrectInput() throws IOException {
         withInput("2.2", inputUtility -> {
             var dble = inputUtility.nextDouble("prompt", d -> true);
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+ * Tests the {@link InputUtility} class.
+ */
+@RunWith(Theories.class)
+public class InputUtilityTest {
+
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
```
nibsi: Annotations that don't take arguments don't need parentheses.

core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Tests the {@link InputUtility} class.
+ */
+@RunWith(Theories.class)
+public class InputUtilityTest {
+
+
+    /**
+     * @throws IOException
+     */
```
nibsi: Either fill out the documentation completely, or remove the stubs.

core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Theory()
+    public void nextString_validatesCorrectInput() throws IOException {
+        var testCase = "input";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
```
nibsi: You can probably simplify a lot of code in this class if you add a utility method:

```java
private static <X> void withInput(
    String input,
    ExceptionalConsumer<? super InputUtility, ? extends X> action
) throws X {
    try (
        var stream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8));
    ) {
        var inputUtility = new InputUtility(stream);
        action.accept(inputUtility);
    }
}
```
```java
@FunctionalInterface
private static interface ExceptionalConsumer<T, X> {

  void accept(T argument) throws X;
  
}
```
```java
withInput("input", inputUtility -> {
    var string = inputUtililty.nextString("prompt");
    assertThat("Next string must match expected value.", string, is("input"));
});
```

core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * @throws IOException
+     */
+    @Theory()
+    public void nextInt_validatesCorrectInput() throws IOException {
+        var testCase = "1";
+        try(var s = new ByteArrayInputStream(testCase.getBytes())) {
+            var iu = new InputUtility(s);
+            var result =  iu.nextInt("prompt", n -> n == 1);
```
nibsi: Don't use the predicate to make your assertion, that's what the `assertThat` call is for.

core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+            assertThat("Matches testCase", Integer.parseInt(testCase), is(equalTo(result)));
```
nibsi: The second argument must be the actual value and the third argument must be a matcher that describes the expected value.

Don't parse the expected value. Specify it explicitly:

```java
assertThat("Next int must match expected value.", nextInt, is(1));
```

core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+
+    public void testOneOfThese() {
+    }
+
+    public void testIntRange() {
+    }
+
+    public void testDoubleRange() {
+    }
```
nibsi: Empty stubs.

core/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,126 @@
...
+}
```
nibsi: Terminate source files with an empty new line.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collector;
+
+/**
+ * Multipurpose input utility for use with a command line interface.
+ * The user can pass a prompt and optional validation.
+ * The utility will read the provided input stream and return a possibly validated String or primitive.
+ */
+public class InputUtility {
```
nibsi: Make classes final.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+    private static final List<String> NO = Collections.unmodifiableList(List.of("n",  "no"));
+
+    private InputStream inputStream;
+
+    /**
+     * If no input stream is provided, System.in will be used.
+     */
+    public InputUtility() {
+        this(System.in);
+    }
```
nibsi: Personally I think the input stream should always be explicitly specified.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * Constructor that takes an input stream to be used.
+     *
+     * @param inputStream input stream to be used.
+     */
+    public InputUtility(InputStream inputStream) {
```
nibsi: Require the client to pass a `Charset` to read strings with.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * Displays the prompt and reads the input stream, returning a possibly validated String.
+     *
+     * @param prompt the prompt to display to the user.
+     * @param stringPredicate the predicate to use for validation.
+     * @return a possibly validated String.
+     * @throws IOException
+     */
+    public String nextString(String prompt, Predicate<String> stringPredicate) throws IOException {
```
nibsi: A predicate is a consumer, so declare them as `Predicate<? super String>`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+        this.inputStream = inputStream;
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a possibly validated String.
+     *
+     * @param prompt the prompt to display to the user.
+     * @param stringPredicate the predicate to use for validation.
+     * @return a possibly validated String.
+     * @throws IOException
```
nibsi: Specify the conditions when an exception is thrown.

util/src/main/java/com/coderanch/util/cli/package-info.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (C) 2018 Coderanch.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/**
+ * Custom Hamcrest matchers to use in parameter checking or unit tests.
```
nibsi: Incorrect package summary.

nibsi (changes requested):

util/src/main/java/com/coderanch/util/cli/InputUtility.java
ksnortum: But it's nice to have a reasonable default, isn't it?  Especially since most of the time you'll be using System.in, except for testing.

nibsi: I agree with reasonable defaults, as long as reasonable defaults don't lead to environment-specific behavior.

`System.in` is really not much more than a mutable global variable and its behavior depends on operating system, execution profile, and more.

I like to compare it to `String.getBytes()`. This method is deprecated because it uses an environment-specific character set.

Anyway, this is not a blocker for me, if you and Al want to keep it in I won't reject the PR because of it.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    public String nextString(String prompt, Predicate<String> stringPredicate) throws IOException {
+        System.out.println(prompt);
+        try(var br = new BufferedReader(new InputStreamReader(inputStream))){
+            while(true) {
+              var line = br.readLine();
+              if(stringPredicate.test(line)) {
```
ksnortum: Needs a space after `if` and before `(`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+     * Displays the prompt and reads the input stream, returning a possibly validated String.
+     *
+     * @param prompt the prompt to display to the user.
+     * @param stringPredicate the predicate to use for validation.
+     * @return a possibly validated String.
+     * @throws IOException
+     */
+    public String nextString(String prompt, Predicate<String> stringPredicate) throws IOException {
+        System.out.println(prompt);
+        try(var br = new BufferedReader(new InputStreamReader(inputStream))){
```
ksnortum: Needs a space after `try`and before `(`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+            while(true) {
```
ksnortum: Needs a space after `while`and before `(`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
ksnortum: The basic rule is, if it's not a method, then add a space before the `(`, otherwise, no space.  Most IDEs have a Format Code function that will fix this.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+     * @param stringPredicate the predicate to use for validation.
+     * @return true for yes and false for no.
+     * @throws IOException
+     */
+    public boolean nextYesNo(String prompt, Predicate<String> stringPredicate) throws IOException {
+        var result = this.nextString(prompt, yesOrNo()).toLowerCase();
+        if(YES.contains(result)) {
+            return true;
+        }
+        else return false;
```
ksnortum: I believe this can be simplified to `return YES.contains(result);`

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * @throws IOException
+     */
+    public void pause() throws IOException {
+        pause("Press when ready");
+    }
+
+    /**
+     * Calls nextString{@link InputUtility#nextString(String, Predicate)}.
+     * Uses prompt entered and waits for the user to press ".".
```
ksnortum: Why 'press "."' instead of 'press &lt;enter&gt;'?

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Get a String predicate that tests for (case insensitive) "y", "n", "yes", "no".
+     *
+     * @return  a String predicate that tests for (case insensitive) "y", "n", "yes", "no".
+     */
+    public static Predicate<String> yesOrNo() {
+        return s -> {
+            var cleanedString = s.trim().toLowerCase();
```
ksnortum: You're not using `cleanedString`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+            if(YES.contains(s)) {
+                return true;
+            }
+            else if(NO.contains(s)) {
+                return true;
+            }
+            return false;
```
ksnortum: This can be simplified.  What about `return YES.contains(cleanedString) || NO.contains(cleanedString);`?

Beware of situations like `if (condition) { return true; }`.  They can almost always be simplified to `return condition;`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+        };
+    }
+
+    /**
+     * Get a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     *
+     * @param these the strings for the input to match to.
+     * @return a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     */
+    public static Predicate<String> oneOfThese(String...these){
```
ksnortum: I believe there should be a space between `String...` and `these`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+        return s -> {
```
ksnortum: Before this, what about a test to see if more than one String was passed?  Maybe throw an IllegalArgumentException if not.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+            var result = Arrays.stream(these)
+                    .filter(option -> option.equalsIgnoreCase(s))
+                    .findFirst()
```
ksnortum: I believe you can use `anyMatch(predicate)` here.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * Get an int predicate that tests whether the input is between the two limits.
+     *
+     * @param lower the lower bound
+     * @param upper the exclusive upper bound
+     * @return an int predicate that tests whether the input is between the two limits.
+     */
+    public static Predicate<Integer> intRange(int lower, int upper) {
+        return i -> {
```
ksnortum: Maybe before this add a test to check that lower isn't greater than upper?

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+            if(i >= upper) {
+                return false;
```
ksnortum: `if (condition) { return false; }` problem mentioned before.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Get a double predicate that tests whether the input is between the two limits.
+     *
+     * @param lower the lower bound
+     * @param upper the exclusive upper bound
+     * @return a double predicate that tests whether the input is between the two limits.
+     */
+    public static Predicate<Double> doubleRange(double lower, double upper) {
```
ksnortum, line 246: Both comments for `intRange` apply here.

ksnortum (commented):
Good job on this class.  I'm a little nit-picky, but this is what this project is all about.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
aquino-a: Wasn't sure what you wanted. I was confused so I just used the "." at the end of the sentence.

nibsi:
I moved `InputUtilityTest` to the correct module for you. If you made changes in the mean time, please merge them and don't use force-push.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
ksnortum: The final `if` is not needed and confusing.  Just use:

`return YES.contains(cleanedString) || NO.contains(cleanedString);`

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * Multipurpose input utility for use with a command line interface.
+ * The user can pass a prompt and optional validation.
+ * The utility will read the provided input stream and return a possibly validated String or primitive.
+ */
+public final class InputUtility implements Closeable, AutoCloseable {
```
nibsi: `Closeable` extends `AutoCloseable`, so you don't need to declare `AutoCloseable`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * Possible inputs for yes.
```
nibsi: I think the following reads a bit more clearly:
```suggestion
     * Allowed user inputs that are considered synonymous with "yes".
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    private static final List<String> YES = Collections.unmodifiableList(List.of("y", "yes"));
```
nibsi: I'd prefer the constant to be named `YES_SYNONYMS`, but it's not a big deal.

nibsi: `List.of()` already returns an unmodifiable list.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    private static final List<String> NO = Collections.unmodifiableList(List.of("n", "no"));
+
+    /**
+     * The reader to read user responses with.
+     */
+    private final BufferedReader reader;
+
+    /**
+     * If no input stream is provided, System.in will be used.
```
nibsi: Start constructor descriptions with the text `Constructs`.

```suggestion
     * Constructs a new input utility that reads from {@link System#in} using the UTF-8 encoding.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    public InputUtility() {
+        this(System.in, StandardCharsets.UTF_8);
+    }
+
+    /**
+     * Constructor that takes an input stream to be used.
```
nibsi, line 53: Document both constructors that using them will cause the input to be buffered.

This is very important, because it means that you can't wrap the underlying input stream in two different utilities simultaneously.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+     *
+     * @param inputStream input stream to be used.
+     * @param charset     charset to be used.
+     */
+    public InputUtility(InputStream inputStream, Charset charset) {
+        reader = new BufferedReader(new InputStreamReader(inputStream, charset));
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a possibly validated String.
```
nibsi: Why 'possibly'? In what case is the string not validated?

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Constructor that takes an input stream to be used.
+     *
+     * @param inputStream input stream to be used.
+     * @param charset     charset to be used.
+     */
+    public InputUtility(InputStream inputStream, Charset charset) {
+        reader = new BufferedReader(new InputStreamReader(inputStream, charset));
```
nibsi, line 63: Perform parameter validation.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a possibly validated String.
+     *
+     * @param prompt          the prompt to display to the user.
+     * @param stringPredicate the predicate to use for validation.
+     * @return a possibly validated String.
+     * @throws IOException when there's a problem with {@link InputStream}
```
nibsi: What input stream?

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,245 @@
...
+        System.out.println(prompt);
+        while (true) {
+            var line = reader.readLine();
+            if (stringPredicate.test(line)) {
+                return line;
+            }
+            else {
+                System.out.println("Invalid input.");
+            }
+        }
```
nibsi: Don't use `true` and `false` as loop conditions.
```suggestion
        String line;
        
        for (
            line = reader.readLine();
            !stringPredicate.test(line);
            line = reader.readLine();
        ) {
            System.out.println("Invalid input.");
        }
        
        return line;
```

nibsi (changes requested):

util/src/main/java/com/coderanch/util/cli/InputUtility.java
ksnortum: That's verbiage from my issue I'm afraid. What I meant there was that there should be overloaded methods, some that have validation and some that don't. But I agree that a method with a predicate for a parameter will always validate.

aquino-a: >  don't use force-push.
@nibsi You mean for this specific case, right?  I use force push sometimes if something minor was missing from the last commit.

aquino-a:
>  don't use force-push.
@nibsi You mean for this specific case, right?  I use force push sometimes if something minor was missing from the last commit.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+import static com.coderanch.util.require.Require.requireThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+/**
+ * Multipurpose input utility for use with a command line interface.
+ * The user can pass a prompt and optional validation.
+ * The utility will read the provided input stream and return a possibly validated String or primitive.
+ */
+public final class InputUtility implements AutoCloseable {
```
nibsi: Implementing `Closeable` was correct. `AutoCloseable` is for resources that aren't related to I/O.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+    private static final List<String> NO_SYNONYMS = List.of("n", "no");
+
+    /**
+     * The reader to read user responses with.
+     */
+    private final BufferedReader reader;
+
+    /**
+     * Constructs a new input utility that reads from {@link System#in} using the UTF-8 encoding.
+     * This constructor will buffer the {@link InputStream}.
```
nibsi:
```suggestion
     * Data from the underlying stream will be buffered.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+     * You cannot use the same stream in a different utility.
```
nibsi:
```suggestion
     * Do not use the same stream in a different utility.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+     */
+    public InputUtility() {
+        this(System.in, StandardCharsets.UTF_8);
+    }
+
+    /**
+     * Constructor that takes an input stream to be used.
+     * This constructor will buffer the {@code inputStream}.
```
nibsi:
```suggestion
     * Data from the underlying stream will be buffered.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+     * You cannot use the same stream in a different utility.
```
nibsi:
```suggestion
     * Do not use the same stream in a different utility.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        requireThat("charset", charset, is(notNullValue()));
+        reader = new BufferedReader(new InputStreamReader(inputStream, charset));
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated String.
+     *
+     * @param prompt          the prompt to display to the user.
+     * @param stringPredicate the predicate to use for validation.
+     * @return a possibly validated String.
```
nibsi:
```suggestion
     * @return a validated string.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+     * @throws IOException when there's a problem with the underlying {@link BufferedReader}.
```
nibsi:
```suggestion
     * @throws IOException if there's a problem while reading from the underlying stream.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+     */
+    public String nextString(String prompt, Predicate<? super String> stringPredicate) throws IOException {
+        System.out.println(prompt);
+        String line;
+        for (
+                line = reader.readLine();
```
nibsi: Continuation size.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated integer.
+     *
+     * @param prompt       the prompt to display to the user.
+     * @param intPredicate the predicate to use for validation.
+     * @return a possibly validated integer.
+     * @throws IOException when there's a problem with {@link InputStream}
+     */
+    public int nextInt(String prompt, Predicate<Integer> intPredicate) throws IOException {
```
nibsi: `Predicate` should not have an upper bound.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        System.out.println(prompt);
+        String line;
+        Integer num;
+        for (
+                line = reader.readLine();
+                (num = this.tryIntParse(line, intPredicate)) == null;
```
nibsi: Don't use nested assignments.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    private Integer tryIntParse(String line, Predicate<Integer> intPredicate) {
+        try {
+            var num = Integer.parseInt(line);
+            if (intPredicate.test(num)) {
+                return num;
+            }
+            else {
+                return null;
```
nibsi: Never return `null`. Use `OptionalInt`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+                (num = this.tryIntParse(line, intPredicate)) == null;
+                line = reader.readLine()
+        ) {
+            System.out.println("Invalid input.");
+        }
+
+        return num;
+    }
+
+    private Integer tryIntParse(String line, Predicate<Integer> intPredicate) {
```
nibsi: Missing documentation.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        try {
+            var num = Integer.parseInt(line);
+            if (intPredicate.test(num)) {
```
nibsi: Use a ternary expression.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+                return num;
+            }
+            else {
+                return null;
+            }
+        }
+        catch (Exception e) {
```
nibsi: Catch more specific exception types.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated real number.
+     *
+     * @param prompt          the prompt to display to the user.
+     * @param doublePredicate the predicate to use for validation.
+     * @return a possibly validated real number.
```
nibsi: Doubles can't represent real numbers, only a subset of rational numbers.

```suggestion
     * @return a validated floating point number.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+     * @throws IOException when there's a problem with {@link InputStream}
```
nibsi: Same comments apply to this method that applied to the previous method.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+                return null;
+            }
+        }
+        catch (Exception e) {
+            return null;
+        }
+    }
+
+    /**
+     * Calls {@link InputUtility#nextString(String, Predicate)}.
```
nibsi: Implementation details must not be the first line of a method summary.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        };
+    }
+
+    /**
+     * Get a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     *
+     * @param these the strings for the input to match to.
+     * @return a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     */
+    public static Predicate<? super String> oneOfThese(String... these) {
```
nibsi: Don't use wildcards in return types.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+    public static Predicate<Integer> intRange(int lower, int upper) {
+        if (lower > upper) {
+            throw new IllegalArgumentException("Lower must be less than upper.");
+        }
+        return i -> {
+            if (i >= upper) {
+                return false;
+            }
+            return i >= lower;
+        };
```
nibsi:
```suggestion
        return i -> lower <= i && i < upper;
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+    public static Predicate<Double> doubleRange(double lower, double upper) {
+        if (lower > upper) {
+            throw new IllegalArgumentException("Lower must be less than upper.");
+        }
+        return d -> {
+            if (d >= upper) {
+                return false;
+            }
+            return !(d < lower);
+        };
```
nibsi:
```suggestion
        return d -> lower <= d && d < upper;
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void close() throws IOException {
+        if (this.reader != null) {
```
nibsi: Is `reader` ever null?

nibsi (changes requested):

util/src/main/java/com/coderanch/util/cli/InputUtility.java
aquino-a: You mean want `Predicate<Number>` ?

nibsi: No, I mean `Predicate<? super Integer>`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,277 @@
...
+        for (
+            line = reader.readLine(), num = OptionalInt.empty();
+            num.isEmpty();
+            line = reader.readLine()
+        ) {
+            num = tryIntParse(line, intPredicate);
+            if (num.isEmpty()) {
+                System.out.println("Invalid input.");
+            }
+        }
```
nibsi:
```suggestion
        String line;
        OptionalInt num;
        for (
            line = reader.readLine(), num = tryIntParse(line, intPredicate);
            num.isEmpty();
            line = reader.readLine(), num = tryIntParse(line, intPredicate)
        ) {
            System.out.println("Invalid input.");
        }
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    /**
+     * Get a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     *
+     * @param these the strings for the input to match to.
+     * @return a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     */
+    public static Predicate<String> oneOfThese(String... these) {
+        if (these.length < 2) {
```
nibsi: Why would you need at least two? One seems reasonable. Also, enforce a minimum number of arguments by detaching them from the varargs parameter:

`oneOfThese(String firstOption, String... otherOptions)`

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Get an int predicate that tests whether the input is between the two limits.
+     *
+     * @param lower the lower bound
+     * @param upper the exclusive upper bound
+     * @return an int predicate that tests whether the input is between the two limits.
+     */
+    public static Predicate<Integer> intRange(int lower, int upper) {
+        if (lower > upper) {
```
nibsi: What would happen if you created and used a predicate where the lower bound is equal to the upper bound?

util/src/test/java/com/coderanch/util/cli/InputUtilityTest.java
```diff
@@ -0,0 +1,193 @@
...
+        var result = InputUtility.doubleRange(values[0], values[1]).test(values[0] + 1);
+        assertThat("Predicate must match expected value.", result, is(true));
+    }
+
+    private static <X extends Throwable> void withInput(
+            String input,
+            ExceptionalConsumer<? super InputUtility, X> action
+    ) throws IOException, X {
+        try (
+                var stream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8));
```
nibsi: More indentation issues.

nibsi (changes requested):

util/src/main/java/com/coderanch/util/cli/InputUtility.java
aquino-a: two was from the original requirements

util/src/main/java/com/coderanch/util/cli/InputUtility.java
aquino-a: it would never be true

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+        return line;
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated integer.
+     *
+     * @param prompt       the prompt to display to the user.
+     * @param intPredicate the predicate to use for validation.
+     * @return a validated integer.
+     * @throws IOException when there's a problem with {@link InputStream}
```
nibsi: Make all exceptions descriptions consistent with the exception description of the `nextString()` method.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+            num.isEmpty();
+            line = reader.readLine(), num = tryIntParse(line, intPredicate)
+        ) {
+            System.out.println("Invalid input.");
+        }
+        return num.getAsInt();
+    }
+
+    /**
+     * Tries to parse an Integer from the string and validate it.
```
nibsi:
```suggestion
     * Tries to parse an integer from the specified line and validates it.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+        requireThat("charset", charset, is(notNullValue()));
+        reader = new BufferedReader(new InputStreamReader(inputStream, charset));
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated String.
+     *
+     * @param prompt          the prompt to display to the user.
+     * @param stringPredicate the predicate to use for validation.
+     * @return a validated string.
```
nibsi:
```suggestion
     * @return a string that's valid according to {@code stringPredicate}.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+
+        return line;
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated integer.
+     *
+     * @param prompt       the prompt to display to the user.
+     * @param intPredicate the predicate to use for validation.
+     * @return a validated integer.
```
nibsi:
```suggestion
     * @return an integer that's valid according to {@code intPredicate}.
```
Make similar changes to the following methods.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+            System.out.println("Invalid input.");
+        }
+        return num.getAsInt();
+    }
+
+    /**
+     * Tries to parse an Integer from the string and validate it.
+     *
+     * @param line         the string to parse.
+     * @param intPredicate the predicate used to validate the Integer.
```
nibsi: Write prose unless you have a good reason to refer to the Java type specifically.
```suggestion
     * @param intPredicate the predicate used to validate the integer.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+            var num = Integer.parseInt(line);
+            return intPredicate.test(num) ? OptionalInt.of(num) : OptionalInt.empty();
+        }
+        catch (NumberFormatException e) {
+            return OptionalInt.empty();
+        }
+    }
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated real number.
```
nibsi:
```suggestion
     * Displays the prompt and reads the input stream, returning a validated floating point number.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+            }
+        }
+        return num.getAsDouble();
+    }
+
+    /**
+     * Tries to parse a Double from the string and validate it.
+     *
+     * @param line            the string to parse.
+     * @param doublePredicate the predicate used to validate the Integer.
```
nibsi: Copy-pasted description.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+        return num.getAsInt();
+    }
+
+    /**
+     * Tries to parse an Integer from the string and validate it.
+     *
+     * @param line         the string to parse.
+     * @param intPredicate the predicate used to validate the Integer.
+     * @return a validated Integer or nothing.
```
nibsi:
```suggestion
     * @return an integer parsed from {@code line} if it's valid according to {@code intPredicate};
     *         {@code Optional.empty()} otherwise.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+     *
+     * @throws IOException when there's a problem with {@link InputStream}
+     */
+    public void pause() throws IOException {
+        pause("Press when ready");
+    }
+
+    /**
+     * Calls nextString{@link InputUtility#nextString(String, Predicate)}.
+     * Uses prompt entered and waits for the user to press <enter>.
```
nibsi: Remember, Javadoc includes HTML syntax, so you need to escape special characters.
```suggestion
     * Uses prompt entered and waits for the user to press &lt;enter&gt;.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+     * @return true for yes and false for no.
+     * @throws IOException when there's a problem with {@link InputStream}
+     */
+    public boolean nextYesNo(String prompt) throws IOException {
+        var result = this.nextString(prompt, yesOrNo()).trim().toLowerCase();
+        return YES_SYNONYMS.contains(result);
+    }
+
+    /**
+     * Calls {@link InputUtility#pause(String)}.
```
nibsi: Don't use implementation details as the first line of a method summary.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+     * The prompt defaults to "Press when ready".
+     *
+     * @throws IOException when there's a problem with {@link InputStream}
+     */
+    public void pause() throws IOException {
+        pause("Press when ready");
```
nibsi: Press what?

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Gets the next yes or no answer.
+     * Calls {@link InputUtility#nextString(String, Predicate)}.
+     * Passes the prompt and a predicate that validates "y", "n", "yes", "no".
+     *
+     * @param prompt the prompt to display to the user.
+     * @return true for yes and false for no.
+     * @throws IOException when there's a problem with {@link InputStream}
```
nibsi: Replace all occurrences of `when` in exception descriptions with `if`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+    public static Predicate<String> yesOrNo() {
+        return s -> {
+            var cleanedString = s.trim().toLowerCase();
+            return YES_SYNONYMS.contains(cleanedString) || NO_SYNONYMS.contains(cleanedString);
+        };
+    }
+
+
+    /**
+     * Get a String predicate that takes two or more Strings and tests whether any one of them match the input.
```
nibsi: "two or more" is no longer correct.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,267 @@
...
+     * Get an int predicate that tests whether the input is between the two limits.
+     *
+     * @param lower the lower bound
+     * @param upper the exclusive upper bound
+     * @return an int predicate that tests whether the input is between the two limits.
+     */
+    public static Predicate<Integer> intRange(int lower, int upper) {
+        if (lower >= upper) {
+            throw new IllegalArgumentException("Lower must be less than upper.");
+        }
```
nibsi: Use `requireThat` with the appropriate [Hamcrest matcher](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html).

nibsi (changes requested):
Some of my comments refer to more than one method. Check all your changes.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -18,8 +18,7 @@
 import java.util.stream.Stream;
 
 import static com.coderanch.util.require.Require.requireThat;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.*;
```
nibsi: Use star imports only for the standard API.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -177,29 +176,30 @@ private OptionalDouble tryDoubleParse(String line, Predicate<? super Double> dou
...
      */
     public boolean nextYesNo(String prompt) throws IOException {
         var result = this.nextString(prompt, yesOrNo()).trim().toLowerCase();
         return YES_SYNONYMS.contains(result);
     }
 
     /**
-     * Calls {@link InputUtility#pause(String)}.
+     * Waits for the user to press &lt;enter&gt;.
      * The prompt defaults to "Press when ready".
```
nibsi: Documentation is out of date.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -177,29 +176,30 @@ private OptionalDouble tryDoubleParse(String line, Predicate<? super Double> dou
...
+     * Calls {@link InputUtility#pause(String)}.
```
nibsi: This seems like unnecessary information to me.

nibsi: Yeah, sorry for the trouble, but please just remove it from all methods.

Documenting which methods call which other methods of the same class is good practice if and only if the class is non-final and designed for inheritance. That way, the programmer will know what effect overriding a method will have.

Since this class is final, you can just get rid off all documentation of self-use.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -238,9 +238,7 @@ public void pause(String prompt) throws IOException {
      * @return an int predicate that tests whether the input is between the two limits.
      */
     public static Predicate<Integer> intRange(int lower, int upper) {
-        if (lower >= upper) {
-            throw new IllegalArgumentException("Lower must be less than upper.");
-        }
+        requireThat("lower", lower, is(lessThan(upper)));
```
nibsi: Good job, but I'd prefer it if you validated `upper` based on `lower`, rather than the other way around.

The reason for this is that `lower` is the first parameter, and as such no parameters come before it that it is constrained by. `upper` comes afterwards, and it makes sense to constrain it based on the parameter that came before it.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -252,9 +250,7 @@ public void pause(String prompt) throws IOException {
      * @return a double predicate that tests whether the input is between the two limits.
      */
     public static Predicate<Double> doubleRange(double lower, double upper) {
-        if (lower >= upper) {
-            throw new IllegalArgumentException("Lower must be less than upper.");
-        }
+        requireThat("lower", lower, is(lessThan(upper)));
```
nibsi: Same here.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+            num = tryDoubleParse(line, doublePredicate);
+            if (num.isEmpty()) {
+                System.out.println("Invalid input.");
+            }
+        }
+        return num.getAsDouble();
+    }
+
+    /**
+     * Tries to parse a Double from the string and validate it.
```
nibsi:
```suggestion
     * Tries to parse a floating point number from the string and validates it.
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * @param line            the string to parse.
+     * @param doublePredicate the predicate used to validate the double.
+     * @return a validated Double or nothing.
```
nibsi: Use documentation that's consistent with `tryIntParse`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+     * Gets the next yes or no answer.
+     * Calls {@link InputUtility#nextString(String, Predicate)}.
+     * Passes the prompt and a predicate that validates "y", "n", "yes", "no".
+     *
+     * @param prompt the prompt to display to the user.
+     * @return true for yes and false for no.
+     * @throws IOException if there's a problem while reading from the underlying stream.
+     */
+    public boolean nextYesNo(String prompt) throws IOException {
+        var result = this.nextString(prompt, yesOrNo()).trim().toLowerCase();
```
nibsi: Always explicitly pass a locale to string operations.

```suggestion
        var result = this.nextString(prompt, yesOrNo()).trim().toLowerCase(Locale.ENGLISH);
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Uses prompt entered and waits for the user to press &lt;enter&gt;.
+     * Calls nextString{@link InputUtility#nextString(String, Predicate)}.
+     *
+     * @param prompt the prompt to display to the user.
+     * @throws IOException if there's a problem while reading from the underlying stream.
+     */
+    public void pause(String prompt) throws IOException {
+        this.nextString(prompt, String::isBlank);
```
nibsi: I wouldn't force a blank line. If there's garbage input on the line, that doesn't mean that the user didn't press enter, which is all we care about.

Normally I'd say, call a method overloads that doesn't accept a predicate and is just always happy, but this PR is getting kinda long so we can fix that in the future.

For now, just pass a predicate that returns true.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /**
+     * Get a String predicate that tests for (case insensitive) "y", "n", "yes", "no".
+     *
+     * @return a String predicate that tests for (case insensitive) "y", "n", "yes", "no".
+     */
+    public static Predicate<String> yesOrNo() {
+        return s -> {
+            var cleanedString = s.trim().toLowerCase();
```
nibsi: Pass locale explicitly.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
nibsi: Not resolved.

Also, please write `string` in lower case.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+            return YES_SYNONYMS.contains(cleanedString) || NO_SYNONYMS.contains(cleanedString);
+        };
+    }
+
+
+    /**
+     * Get a String predicate that takes two or more Strings and tests whether any one of them match the input.
+     *
+     * @param firstOption the first of the strings to match to.
+     * @param otherOptions the strings for the input to match to.
```
nibsi: In all methods in this class that have a varargs parameter, please rewrite the description so that it refers to "the other strings" instead of just "the strings".

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+     * @return a String predicate that takes one or more Strings and tests whether any one of them match the input.
+     */
+    public static Predicate<String> oneOfThese(String firstOption, String... otherOptions) {
+        return s -> Stream.concat(Arrays.stream(otherOptions), Stream.of(firstOption))
+            .anyMatch(choice -> choice.equalsIgnoreCase(s.trim()));
```
nibsi: It doesn't really affect the outcome of the method, but I find it strange that you would  concat the first option to the end of the other options, rather than the other way around.

nibsi: There's no mention that this method is case insensitive.

I'm also not sure if it's the correct approach. What if the client DOES what case sensitivity?

util/src/main/java/com/coderanch/util/cli/InputUtility.java
nibsi: Replace all occurrences of `trim()` with `strip()`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /**
+     * Get an int predicate that tests whether the input is between the two limits.
```
nibsi: Start method descriptions with a verb in present tense, not imperative mood:

```suggestion
     * Gets an int predicate that tests whether the input is between the two limits.
```

Fix this in other methods as well.

nibsi (changes requested):

util/src/main/java/com/coderanch/util/cli/InputUtility.java
aquino-a: If the user is choosing an option then there's no reason to be case sensitive, because the meaning of words won't change based on case. I can remove it or add an option if you want.

nibsi: Well, what if the option isn't a word, but a symbol? What if the application has to do one thing when the user enters "x" but another thing when the user enters "X"?

Remember, this is a general purpose library, and we can't assume it won't be used for more than just Blackjack.

You don't have to add or remove anything, just rename the method to `oneOfTheseIgnoringCase`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,262 @@
...
+            return doublePredicate.test(num) ? OptionalDouble.of(num) : OptionalDouble.empty();
+        }
+        catch (NumberFormatException e) {
+            return OptionalDouble.empty();
+        }
+    }
+
+    /**
+     * Gets the next yes or no answer.
+     * Passes the prompt and a predicate that validates "y", "n", "yes", "no".
```
nibsi, line 169: Add a comment saying that this method is case insensitive.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated integer.
+     *
+     * @param prompt       the prompt to display to the user.
+     * @param intPredicate the predicate to use for validation.
+     * @return an integer that's valid according to {@code intPredicate}.
+     * @throws IOException if there's a problem while reading from the underlying stream.
+     */
+    public int nextInt(String prompt, Predicate<? super Integer> intPredicate) throws IOException {
```
nibsi: Sorry, this should be an `IntPredicate` instead of a `Predicate<? super Integer>`.

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Displays the prompt and reads the input stream, returning a validated floating point number.
+     *
+     * @param prompt          the prompt to display to the user.
+     * @param doublePredicate the predicate to use for validation.
+     * @return a validated floating point number.
+     * @throws IOException if there's a problem while reading from the underlying stream.
+     */
+    public double nextDouble(String prompt, Predicate<? super Double> doublePredicate) throws IOException {
```
nibsi: And this should be a `DoublePredicate`.

nibsi (changes requested):
Almost there!

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -11,6 +11,8 @@
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
+import java.util.function.DoublePredicate;
+import java.util.function.IntPredicate;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
```
nibsi:
```suggestion
import java.nio.charset.*;
import java.util.*;
import java.util.function.*;
import java.util.stream.*;
```

util/src/main/java/com/coderanch/util/cli/InputUtility.java
```diff
@@ -172,6 +174,7 @@ private OptionalDouble tryDoubleParse(String line, Predicate<? super Double> dou
     /**
      * Gets the next yes or no answer.
      * Passes the prompt and a predicate that validates "y", "n", "yes", "no".
+     * Is case insensitive.
```
nibsi:
```suggestion
     * This method is case insensitive.
```

nibsi:
Don't forget to merge the master branch into this branch.

nibsi (dismissed):

nibsi:
@ksnortum do you want to give this a final review?

ksnortum:
Tomorrow, if I remember!

nibsi (approved):

ksnortum (approved):
Looks good!

PR #32, opened 2021-03-19
Main application loop to build a hand of cards

### LINKED ISSUE #25: Write main application loop to build a hand of cards
- Add main method to `Hand`.
- Repeatedly ask user to "hit" or "pass".
- If the user hits, another card must be added to their hand. The new hand is displayed.
- If the hand goes bust after hitting, a message is displayed to the user and the application terminates.
- If the user passes, the application terminates.

Depends on #21

aquino-a:
resolves #25


aquino-a:
Does the main application really need code coverage?

nibsi:
> Does the main application really need code coverage?

What makes it so special that it doesn't need testing?

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+     *
+     * @return the best score.
+     */
+    public int bestScore() {
+        var scores = calculateScores(this.cards);
+        return calculateBestScore(scores);
+    }
+
+    /**
+     * Issue #25.
```
nibsi: Don't refer to the issue tracker from code. The code must stand on its own, regardless of which repository it lives in what application lifecycle management system is used.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+     *
+     * @param args main arguments
+     */
+    public static void main(String[] args) {
+        try (var iu = new InputUtility()) {
```
nibsi: Don't use abbreviations for variable names. It costs you absolutely nothing to write `input` or `inputUtility`.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+        catch (IOException e) {
+            System.out.println(e);
+            System.exit(1);
+        }
+    }
+
+    /**
+     * Mini game of Blackjack for issue #25.
+     */
+    private static final class MiniGame {
```
nibsi: You should only use internal classes when its obvious that the internal class acts as a component to the outer class.

A `MiniGame` is not a component of a `Hand` of cards, so put it in a separate source file. If you do, it's also better to move the `main` method from the `Hand` class to your new `MiniGame`.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+
+        /**
+         * The player's current hand.
+         */
+        private Hand hand;
+
+        /**
+         * The predicate used for validating user input.
+         */
+        private Predicate<? super String> playPredicate = InputUtility.oneOfThese("hit", "pass");
```
nibsi: Why is it not static and final?

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+
+        /**
+         * The deck of cards used in the game.
+         */
+        private Deque<Card> deck = new ArrayDeque<>(Cards.getShuffledStandardDeck(new SecureRandom()));
```
nibsi: Why is it not final?

nibsi: Why do you use `SecureRandom` instead of `Random`?

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+
+        /**
+         * The input utility used for getting answers.
+         */
+        private InputUtility inputUtility;
+
+        private MiniGame(InputUtility inputUtility) {
+            this.inputUtility = inputUtility;
```
nibsi: When visibility of this constructor goes up, validate the parameter.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        /**
+         * Start the mini game.
```
nibsi: The name `start` implies that a process is triggered and performed in the background, while the method call returns immediately.

Rename it to `run`.

nibsi: Use a better method summary. Explain what it means to "run a game".

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+         *
+         * @throws IOException from inputUtility.
```
nibsi: This description means nothing to a client of this class. Use proper prose:

```
* @throws IOException if an I/O exception occurred while prompting the user for an action.
```

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+         */
+        private void start() throws IOException {
+            hand = new Hand(deck.removeLast(), deck.removeLast());
+            while (true) {
```
nibsi: Don't use `true` and `false` as loop conditions.

```java
displayStatus();

while (!isGameOver()) {
    var choice = ...
    
    switch (choice) {
        case HIT: {
            dealCard();
            displayStatus();
            break;
        }
        
        case PASS: {
            pass();
            break;
        }
    }
}
```
The call to `pass()` should affect the result of `isGameOver()`.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -0,0 +1,199 @@
...
+                if ("pass".equalsIgnoreCase(answer.trim())) {
+                    System.out.println("You passed.  Game over.");
+                    break;
+                }
+                dealCard();
+            }
+        }
+
+        private void displayStatus() {
+            System.out.println("Your cards are: ");
```
nibsi: Don't call `System.out` directly.

Let your game constructor take an instance of `PrintStream`.

nibsi (changes requested):

core/src/main/java/com/coderanch/blackjack/Hand.java
aquino-a: I used `Random()` at first but it came up as a security problem in sonar

nibsi:
Please reopen this PR when the `InputUtility` and the `Hand` PRs have merged. It's getting a bit muddy here.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -113,4 +114,13 @@ private static int calculateBestScore(int minimumScore, int numberOfFreeAces) {
     boolean isBust() {
         return bestScore() > MAX_LEGAL_SCORE;
     }
+
+    /**
+     * Gets the cards of the hand.
+     *
+     * @return an {@code unmodifiableList} of cards.
```
nibsi: Use prose instead of code, unless you are referring to literals or method parameters.

`@return an unmodifiable list of cards.`

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.function.Predicate;
+
+import static com.coderanch.util.require.Require.requireThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+/**
+ * Mini game of Blackjack.
+ */
+public final class MiniGame {
```
nibsi: Why is the class public?

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Highest score in Blackjack.
+     */
+    private static final int MAX_LEGAL_SCORE = 21;
```
nibsi: Code duplication.

You probably don't need this constant because `Hand` has an `isBust()` method.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * The predicate used for validating user input.
+     */
+    private static final Predicate<? super String> PLAY_PREDICATE = InputUtility.oneOfTheseIgnoringCase("hit", "pass");
```
nibsi: I'd just inline this predicate in the line where it is used, because it reads more fluently:

`inputUtility.nextString("hit or pass?", oneOfTheseIgnoringCase("hit", "pass"))`

You can then remove this constant.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * The player's current hand.
+     */
+    private Hand hand;
+
+    /**
+     * The deck of cards used in the game.
+     */
+    private final Deque<Card> deck = new ArrayDeque<>(Cards.getShuffledStandardDeck(new SecureRandom()));
```
nibsi: Don't create services in classes that depend on those services. Inject them into a constructor.

```java
this.inputUtility = requireThat("inputUtility", inputUtility, is(notNullValue()));
this.printStream  = requireThat("printStream",  printStream,  is(notNullValue()));

requireThat("generator", generator, is(notNullValue()));

this.deck = new ArrayDeque<>(Cards.getShuffledStandardDeck(generator));
```
Another thing, why are you using `SecureRandom` and not just `Random`?

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * The input utility used for getting answers.
+     */
+    private final InputUtility inputUtility;
+
+    /**
+     * The print stream used for sending text to the user.
+     */
+    private final PrintStream printStream;
```
nibsi: Prefer `PrintWriter` over `PrintStream`.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Run the mini game.
+     * Will start a game of Blackjack.
+     * Will ask the user to hit or pass until he wins, loses, or passes.
+     *
+     * @throws IOException if an I/O exception occurred while prompting the user for an action.
+     */
+    public void run() throws IOException {
+        hand = new Hand(deck.removeLast(), deck.removeLast());
```
nibsi: `removeLast` is analogous to taking a card from the bottom of a deck. Why not use `removeFirst` or just `remove`?

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * The predicate used for validating user input.
+     */
+    private static final Predicate<? super String> PLAY_PREDICATE = InputUtility.oneOfTheseIgnoringCase("hit", "pass");
+
+    /**
+     * The possible choices the user can make.
+     */
+    private enum CHOICE {
```
nibsi: Type names are never in ALL_CAPS. Rename to `Choice`.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Get the user's next choice.
+     *
+     * @return the user's choice.
+     * @throws IOException if there's a problem with the underlying stream.
+     */
+    private CHOICE getUserChoice() throws IOException {
+        var result = inputUtility.nextString("hit or pass?", PLAY_PREDICATE);
+        return Enum.valueOf(CHOICE.class, result.trim().toUpperCase(Locale.ENGLISH));
```
nibsi: `return Choice.valueOf(result.trim().toUpperCase(Locale.ENGLISH));`

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * The print stream used for sending text to the user.
+     */
+    private final PrintStream printStream;
+
+    /**
+     * Represents whether the user has passed or not.
+     */
+    private boolean isPassed;
```
nibsi: This name implies that the "game is passed", whatever that means.

Rename it to `hasPlayerPassed`.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+        printStream.println(String.format("Your score is: %d", hand.bestScore()));
+    }
+
+    /**
+     * Checks to see if the game is over.
+     *
+     * @return if the game is over or not.
+     */
+    private boolean isGameOver() {
+        if (hand.bestScore() == MAX_LEGAL_SCORE) {
```
nibsi: Add a method `isBlackjack()` to `Hand`.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+            printStream.println("You win. Game over.");
+            return true;
+        }
+        if (hand.isBust()) {
+            printStream.println("You lose. Game over.");
```
nibsi: I don't really expect printing statements in a method whose main purpose is to return a simple property of the object.

Move them to the `run()` method.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+
+
+    /**
+     * Play basic game of Blackjack through the console.
+     *
+     * @param args main arguments
+     */
+    public static void main(String[] args) {
+        try (var inputUtility = new InputUtility()) {
+            var miniGame = new MiniGame(inputUtility, System.out);
```
nibsi: You can create a `PrintWriter` from `System.out` like this:
```java
var writer = new PrintWriter(System.out, true, StandardCharsets.UTF_8);
```

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -0,0 +1,186 @@
...
+            miniGame.run();
+        }
+        catch (IOException e) {
+            System.out.println(e);
```
nibsi: Don't catch the exception. Just declare that the main method throws it.

core/src/test/java/com/coderanch/blackjack/MiniGameTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+
+/**
+ * Tests the {@link MiniGame} class.
+ */
+public class MiniGameTest {
+
```
nibsi, line 27: I'll review the test class after you've updated the code.

nibsi (changes requested):

core/src/main/java/com/coderanch/blackjack/MiniGame.java
aquino-a: I did that in reaction to something you said before. I don't know what it was.

nibsi: https://github.com/CoderanchCorral/Blackjack/pull/32#discussion_r597488245

According to our previous discussion, it was Sonar that flagged this issue. If the warning appears again, I will disable it because the decision to use `Random` or `SecureRandom` is one that should be made by a developer, not an automated tool.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -76,7 +76,7 @@ Hand withAdditionalCard(Card card) {
      * Gets the best score.
      *
      * @return the highest score that can be made by using a large point value for aces
-     *         without the hand going bust, or the lowest score if the hand is bust.
+     *     without the hand going bust, or the lowest score if the hand is bust.
```
nibsi, line 79: While I prefer that lines of Javadoc line up with each other the way they were before your change, I don't prefer it so strongly that we MUST do it my way. The most important thing to me is that we do it consistently across the entire code base.

However, try to avoid making changes to the code that are not related to the feature you want to implement with your pull request. If you want to clean up code, either do it in only the lines that would have been affected by your pull request anyway, or create a separate pull request that is meant only for code cleanup.

You can leave this change as is, it's just a general tip from me.

core/src/main/java/com/coderanch/blackjack/Hand.java
```diff
@@ -109,16 +109,28 @@ private static int calculateBestScore(int minimumScore, int numberOfFreeAces) {
...
         return bestScore() > MAX_LEGAL_SCORE;
     }
 
+    /**
+     * Gets whether the hand is blackjack or not.
+     * A blackjack hand has a score equal to the legal maximum.
+     *
+     * @return {@code true} if {@link #bestScore()} is equal to the maximum legal score;
+     *     {@code false} otherwise.
+     */
```
nibsi, line 124: Very nice.

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -103,36 +96,44 @@ public void run() throws IOException {
                 case PASS:
                 default:
                     pass();
+                    printWriter.println("You passed. Game over.");
```
nibsi, line 98: I'm not a fan of the `PASS` case and the `default` case doing the same thing.

`getPlayerChoice()` should never return anything other than `HIT` or `PASS`. That means the `default` case should never occur. Enforce that assumption by throwing an `AssertionError`:

```java
default:
    throw new AssertionError("Unexpected choice.");
```

core/src/main/java/com/coderanch/blackjack/MiniGame.java
```diff
@@ -103,36 +96,44 @@ public void run() throws IOException {
...
     /**
      * Display the status of the game.
      */
     private void displayStatus() {
-        printStream.println("Your cards are: ");
-        hand.cards().forEach(printStream::println);
-        printStream.println(String.format("Your score is: %d", hand.bestScore()));
+        printWriter.println("Your cards are: ");
+        hand.cards().forEach(printWriter::println);
+        printWriter.println(String.format("Your score is: %d", hand.bestScore()));
```
nibsi: Instead of using `println()` in combination with `format()`, just use `printf()`:

```java
printWriter.printf("Your score is: %d%n, hand.bestScore());
```

nibsi (changes requested):
Very good. A few small changes, but I'm quite pleased.

nibsi:
I updated the code analysis rules. There are 4 code smells left, but you don't have to fix them. I want to overhaul and clean up some stuff after this PR has been merged.

core/src/main/java/com/coderanch/blackjack/Hand.java
aquino-a: I think you wrote that one haha

nibsi (approved):

